PR #5, opened 2019-01-23
Proof of concept for a thread-safe LT instance manager

jonathonherbert:
This PR adds an ValidatorInstancePool class, which is responsible for the lifecycle of threads containing validator instances. It maintains a queue of work -- calls to check() add a validation request to the queue, which is then picked up by the next available thread.

It's not really clear why LanguageTool instances are not thread safe! Going by https://github.com/languagetool-org/languagetool/blob/master/languagetool-server/src/main/java/org/languagetool/server/Pipeline.java, it seems to be a side-effect of JLanguageTool having mutable configuration options -- the Pipeline class walls off methods that mutate configuration once it's done, which seems to be adequate for their use case.

Thanks @mbarton for showing me this pattern!


app/services/LanguageToolFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+}
+
+class LanguageTool(underlying: JLanguageTool) {
+  def check(query: CheckQuery): Seq[RuleMatch] = {
+    println(s"Running check on thread -- #${Thread.currentThread().getId} ${Thread.currentThread().getName}")
+    underlying.check(query.text).asScala.map(RuleMatch.fromLT)
+  }
+
+  def reingestRules(implicit ec: ExecutionContext): Unit = {
+    RuleManager.getAll.map(_.foreach(rule => underlying.addRule(PatternRule.toLT(rule))))
```
mbarton: If we have an instance of Language Tool per thread we might be able to reingest rules by sending new ones down the same queue we use for requests?

Although tbh given that changing rules happens much less often than making normal requests you're probably just as well off creating brand new instances and then flushing the old queues over to the new ones

jonathonherbert: Yes, this is a bit of a hangover from the previous PR, which didn't care about threads -- and you're right, stomping the old instances and creating new ones is probably the way to go.

jonathonherbert:
Closed in favour of #4

PR #7, opened 2019-06-18
Remove tokens dir and add to gitignore

jonathonherbert:


sihil:
What generates the tokens directory? Is there not a way to turn off on disk persistence?

jonathonherbert:
@sihil thanks for reminding me -- we no longer persist to disk, so the change to .gitignore was not needed 👍

PR #8, opened 2019-06-18
Remove RuleManager in favour of an interfaced RuleResource

jonathonherbert:
A bit of a bikeshed-y PR to remove an abstraction which wasn't providing much value.

This means that
- it's simpler to swap between resources, e.g. for testing purposes.
- our configuration params are now extracted at application boot, and the surface area for the instantiation of the SheetsRuleResource is much smaller than the entire config object (thanks @sihil for the advice). This isn't a prerequisite for the refactor, but passing properties transparently through a management class was a bit of a smell.


sihil:
Looks like this needs a fair bit of conflict resolving.

jonathonherbert:
@sihil oh bother, yes, as soon as I merged the previous PR -- and simultaneously pointed you this way! This should now be ready for review.

jonathonherbert:
@sihil ah, forgive my ignorance -- I've never used `git mv` before, that's very useful to know 👍

PR #9, opened 2019-06-24
Do a better job parsing incoming regular expressions

jonathonherbert:
~NB: Review #8 before you review this PR, dear reader! It's upstream from this work.~

This PR uses a different Rule class to manage regular expressions and is more tolerant of absences in incoming data, removing the difficulty the previous PatternRule based implementation had in applying regular expressions that spanned several words.

As a result, all but a single rule in our test corpus is applied. In our previous regex ingestions, we were rejecting dozens of rules as a result of these difficulties, and silently failing on many more.


app/model/LTRegexPatternRule.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Matches 'regexp' elements from XML rules against sentences.
+ *
+ * Copied from the LanguageTool package, which doesn't make this class public.
+ */
+class LTRegexPatternRule extends AbstractPatternRule implements RuleMatcher {
```
jonathonherbert: As suggested by @mbarton, this is one way of getting hold of a class which isn't publicly available. If others are preferable please let me know!

app/model/RuleMatch.scala
```diff
@@ -19,7 +19,7 @@ object RuleMatch {
       fromPos = lt.getFromPos,
       toPos = lt.getToPos,
       message = lt.getMessage,
-      shortMessage = Option(lt.getShortMessage).filterNot(_.isEmpty),
+      shortMessage = Some(lt.getMessage),
```
jonathonherbert: shortMessage isn't present in RegexPatternRule, so we defer to message here. We could add it, but until we know more about our modelling requirements I'm not sure there's a reason to worry about it.

app/model/LTRegexPatternRule.java
```diff
@@ -0,0 +1,169 @@
+/* LanguageTool, a natural language style checker
```
mbarton: Ah interesting, LT is under LGPL. Not a lawyer but I believe copying this code into the project would make Typerighter LGPL too.

I don’t think we currently have a licence in the project? Either way, most of the stuff at the G is Apache.

Not the biggest deal but since we want Typerighter to be usable by other orgs maybe worth doing the reflection trick instead?

Happy to catch up in person to explain further, software licensing is a mess and I am far from an expert on it or what the general approach is at the Guardian

jonathonherbert: @mbarton that's a really good catch! TIL: 'For proprietary software, code under the LGPL is usually used in the form of a shared library, so that there is a clear separation between the proprietary and LGPL components.' Let's catch up.

sihil: I agree with this analysis. Copying from LT is best avoided. We could perhaps submit a PR that changes the access on the class if there is a good reason for it to be public.

jonathonherbert:
Copied code can now be replaced by a direct reference to the class once we point to a version than [includes this change.](https://github.com/languagetool-org/languagetool/pull/1662)

jonathonherbert:
@sihil in the mean time, I've updated this PR to use reflection instead of the copied code.

jonathonherbert:
@sihil all done 👍

jonathonherbert:
@mbarton this is blocking and I'd like to get it in for Max before tomorrow -- any chance you could take a look?

sihil:
Apologies for blocking you here. Not sure what setting to change to allow merges in that situation.

jonathonherbert:
@sihil no problem, there's a path to redemption by taking a look at #8 😁

PR #10, opened 2019-07-03
Add a management screen

jonathonherbert:
~NB: Greetings, contributor! Please review #8 before checking out this PR, as it's been rebased from that branch.~ This is now ready for review.

## What's changed

Adds a management interface. At the moment this simply shows the currently loaded rules, and allows the user to trigger a rule refresh.

Screenshot 2019-07-03 at 16 23 28

I've written this in Twirl as I wasn't sure that we required a full-blown SPA given the simplicity of the UI, but if we now consider React to be more idiomatic for Tools projects, that's fine too; it may be that we drop it in later as the project grows.

I should also add that this is my first time with Twirl, so any feedback would be very useful!


conf/application.conf
```diff
@@ -1,4 +1,9 @@
 play.application.loader=AppLoader
+play.filters.headers.contentSecurityPolicy = "default-src 'self' 'unsafe-inline'"
```
jonathonherbert: This was necessary to allow inline styling, which we use to colour the rule pills. I'm happy to discuss if anyone's worried about the security implications.

project/plugins.sbt
```diff
@@ -6,4 +6,6 @@ addSbtPlugin("io.gatling" % "gatling-sbt" % "3.0.0")
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
 
+addSbtPlugin("com.typesafe.sbt" % "sbt-less" % "1.1.2")
```
jonathonherbert: Super stylin'

app/assets/stylesheets/main.less
```diff
@@ -0,0 +1,32 @@
...
+@font-face {
+  font-family: GHGuardianHeadline;
+  src: url('/assets/fonts/GHGuardianHeadline-Medium.ttf') format('woff2'),
+  url('/assets/fonts/GHGuardianHeadline-Medium.woff') format('woff'),
+  url('/assets/fonts/GHGuardianHeadline-Medium.woff') format('truetype');
+  font-weight: bold;
+  font-weight: 700;
+}
+
+@headline-font: GHGuardianHeadline, "Times New Roman", Serif;
```
jonathonherbert: Couldn't resist.

app/controllers/HomeController.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext}
+
+/**
+ * The controller for the index pages.
+ */
+class HomeController(cc: ControllerComponents)(implicit ec: ExecutionContext)  extends AbstractController(cc) {
+  def index() = Action { implicit request: Request[AnyContent] =>
```
mbarton: Does this need user authentication on it?

jonathonherbert: This PR doesn't address authentication, but it probably should now that Typerighter is out in the wild. So -- yes! I'll take a look at some projects, are there any you'd consider to be good to steal from/wildly unidiomatic? (It might be nice to consider an approach that would permit other organisations to use the project, I remember you were doing something like this for the Grid)

mbarton: Ah ok, not a blocker if it doesn't already have auth.

I think Composer will be accessing the Typerighter API directly? If that's the case then the precedent would be to use Panda in [verify mode](https://github.com/guardian/pan-domain-authentication/#to-verify-logins).

However that does make it non-trivial for others to use Typerighter with authentication, as they need to set up Panda and [login.gutools](https://github.com/guardian/login.gutools). We've found that's quite a lot of work for people working on the Grid.

[google-play-auth](https://github.com/guardian/play-googleauth) (as used by Ophan) is slightly simpler to set up but wouldn't have the single-sign-on behaviour of Panda so using it for AJAX requests from Composer might not be appropriate as users couldn't interactively authenticated themselves.

Interesting, the Amazon ALB load balancer can [authenticate users directly](https://aws.amazon.com/blogs/aws/built-in-authentication-in-alb/). This could mean we push out authenticated as a concern from Typerighter itself, essentially making it pluggable for any users based on what AWS support. This woud be at the expense of (a bit) more expensive load balancer compared to the classic ones we are currently using in other tools.

Summarising my opinion:
 - Fastest to implement: [Panda in verify mode](https://github.com/guardian/pan-domain-authentication/#to-verify-logins)
 - Most interesting for the future (including others using Typerighter): [ALB authentication](https://aws.amazon.com/blogs/aws/built-in-authentication-in-alb/)

cc @sihil for his opinion

jonathonherbert: @mbarton This is a really useful and detailed reply, thanks! An ALB-based solution seems like a step forward for projects more generally, and the [pricing doesn't look like it'll be significantly different with the usage we're expecting](https://aws.amazon.com/elasticloadbalancing/pricing/) (low hundreds of users, low tens concurrently). I'll take a look.

mbarton: Great! We have experimented with that strategy as well on Investigations so would be really interested to pair to get it to work

jonathonherbert:
any further opinions, @mbarton, @sihil? Would love to get this and #11 merged.

jonathonherbert:
@mbarton ace. Re: LB, sounds good 👍 -- typerighter doesn't yet talk to the outside world but it will need to soon

PR #11, opened 2019-07-24
Add the ability to add validators at runtime

jonathonherbert:
~Greetings, traveller! Please direct your attention to #10 before reviewing this PR, as this work is downstream 😁~ This is now ready for review.

This PR adds the ability to add validators at runtime. Previously, the validators that were available to a validationPool were bound at the point of instantiation. Now, they're added as and when, after instantiation, on a per-category basis. This lets us use more than one kind of validator per pool, and also means we could potentially mount/unmount validators at runtime as needed.

### DIFF
--- a/app/controllers/RulesController.scala
+++ b/app/controllers/RulesController.scala
@@ -18,7 +18,7 @@ class RulesController(cc: ControllerComponents, validatorPool: ValidatorPool, la
       errorsByCategory = addValidatorToPool(rulesByCategory)
       rules <- validatorPool.getCurrentRules
     } yield {
-      val errors = errorsByCategory.map(_._2).flatten ::: ruleErrors
+      val errors = errorsByCategory.flatMap(_._2) ::: ruleErrors
       val rulesIngested = rulesByCategory.map { _._2.size }.sum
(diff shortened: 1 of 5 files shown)


app/controllers/RulesController.scala
```diff
@@ -10,18 +11,14 @@ import scala.concurrent.{ExecutionContext, Future}
...
-      errorsByCategory <- Future.sequence(
-        rulesByCategory.map { case (category, rules) => {
-          validatorPool.updateConfig(category.name, ValidatorConfig(rules))
-        }}.toList
-      )
+      errorsByCategory = addValidatorToPool(rulesByCategory)
       rules <- validatorPool.getCurrentRules
     } yield {
-      val errors = errorsByCategory.flatten ::: ruleErrors
+      val errors = errorsByCategory.map(_._2).flatten ::: ruleErrors
```
mbarton: Not a big deal but could be `errorsByCategory.flatMap(_._2)`

jonathonherbert:
@mbarton after making that change, strangely github has now dismissed your review as stale! An additional +1 would be 👌

app/controllers/RulesController.scala
jonathonherbert: 👍

PR #12, opened 2019-07-29
Add name-checker matcher

jonathonherbert:
In draft as the wiring for Elasticsearch is yet to be completed.

## What's changed

Adds a validator that finds names in a document to check spelling.

## Implementation details.

Finding names in a document is relatively trivial with the NER functionality available as part of Stanford CoreNLP. This library was available in Java and seems to [perform as well or better than any alternatives ](https://www.analyticsvidhya.com/blog/2017/04/natural-language-processing-made-easy-using-spacy-%E2%80%8Bin-python/)(save, possibly, for Google's [Parsey McParseface](https://ai.googleblog.com/2016/05/announcing-syntaxnet-worlds-most.html), which is a real thing that exists).

Once we've got the names, it fuzzy-matches them against an elasticsearch database of article names (currently drawn from the [wikipedia name list](https://dumps.wikimedia.org/enwiki/latest/)). We use elasticsearch because it does a better job of matching against names than wikipedia's API -- it's better at matching names that are a larger levenshtein distance away. It also feels quite nice that we have more control over the characteristics of the search, but I've no idea whether that will turn out to be relevant.

At this point, we have a list of similar articles names, but no way of knowing what kind of entity they represent. We thus take the top n results and map them against DBpedia to check that they match the `person` ontology. This winnows out non-person articles and redirects, so we can be reasonably sure that the article we're returning will be relevant.

Finally, we send the article title and text back to the caller, who can pick up any additional wikipedia content with the article name.

### Notes for future changes

- It'd be lovely to be able to do this all offline. At the moment, the name search is done locally but we go to DBpedia once we've got matches. It should be perfectly possible to [host a local version](https://joernhees.de/blog/2014/11/10/setting-up-a-local-dbpedia-2014-mirror-with-virtuoso-7-1-0/).
- I didn't spend too much time optimising, so we can imagine scenarios where e.g. the name search pulls up loads of matches that don't represent people, and we get an empty search result, etc. I reasoned that it's probably better to get this out in front of users before trying to work on these kinds of problems, but there's lots of room for improving the implementation.
- We could experiment with different NLP libraries if Stanford proves insufficient.
- Can we fold user feedback into our model somehow?


jonathonherbert:
Closing, as it's very unlikely this will be used in the short- to medium-term, but this is useful within the repository as an example of one application of NER.

PR #13, opened 2019-08-07
Queue validation jobs

jonathonherbert:
Please see #11 before reviewing this PR -- it's upstream from this work.

## What's changed?

At the moment, there's no limit to the number of concurrent validation jobs, and sending large numbers of validations simultaneously can cause boxes to fall over. This PR adds a queue to the `ValidatorPool` to limit the number of validation jobs that can be performed concurrently. We can adjust this figure to establish sensible defaults.

### Implementation details

I tried a few ways of queuing work before settling on the current Akka streams implementation, including rolling my own queue and using Actors to managing distributing work:
- Rolling my own queue worked and tested well, but I couldn't help but feel there was a solution that was more concise and less surprising
- Actors were quite verbose, and that was before I'd implemented backpressure behaviour

The current stream-based implementation seems to work well and is definitely the most concise and arguably the least surprising of the three. I'd be really interested in any feedback.

### DIFF
--- a/app/services/LanguageToolFactory.scala
+++ b/app/services/LanguageToolFactory.scala
@@ -57,7 +57,9 @@ class LanguageTool(category: String, instance: JLanguageTool)(implicit ec: Execu
   def getCategory = category
 
   def check(request: ValidatorRequest): Future[List[RuleMatch]] = {
-    Future.successful(instance.check(request.text).asScala.map(RuleMatch.fromLT).toList)
+    Future {
+      instance.check(request.text).asScala.map(RuleMatch.fromLT).toList
+    }
   }
 
   def getRules: List[Rule] = {
--- a/app/services/LanguageToolFactory.scala
+++ b/app/services/LanguageToolFactory.scala
@@ -7,9 +7,8 @@ import org.languagetool.rules.{Rule => LTRule}
 import org.languagetool.rules.spelling.morfologik.suggestions_ordering.SuggestionsOrdererConfig
 
 import collection.JavaConverters._
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import play.api.Logger
-
 import model.RuleMatch
 import model.Rule
 import org.languagetool.rules.CategoryId
@@ -55,8 +54,8 @@ class LanguageToolFactory(
 class LanguageTool(category: String, instance: JLanguageTool)(implicit ec: ExecutionContext) extends Validator {
   def getCategory = category
 
-  def check(request: ValidatorRequest): List[RuleMatch] = {
-    instance.check(request.text).asScala.map(RuleMatch.fromLT).toList
+  def check(request: ValidatorRequest): Future[List[RuleMatch]] = {
+    Future.successful(instance.check(request.text).asScala.map(RuleMatch.fromLT).toList)
   }
 
   def getRules: List[Rule] = {
--- a/app/services/LanguageToolFactory.scala
+++ b/app/services/LanguageToolFactory.scala
@@ -52,6 +52,8 @@ class LanguageToolFactory(
 }
 
 class LanguageTool(category: String, instance: JLanguageTool)(implicit ec: ExecutionContext) extends Validator {
+  def getId = "language-tool"
+
   def getCategory = category
 
   def check(request: ValidatorRequest): Future[List[RuleMatch]] = {
--- a/app/services/LanguageToolFactory.scala
+++ b/app/services/LanguageToolFactory.scala
@@ -16,7 +16,7 @@ import utils.Validator
 
 class LanguageToolFactory(
                            maybeLanguageModelDir: Option[File],
-                           useLanguageModelRules: Boolean = false) extends ValidatorFactory {
+                           useLanguageModelRules: Boolean = false) {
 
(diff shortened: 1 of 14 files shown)


test/scala/services/ValidatorPoolTest.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  def check(request: ValidatorRequest) = {
+    val promise = Promise[ValidatorResponse]()
+    val future = promise.future
+    currentWork = Some(promise)
+    maybeComplete()
+    future
+  }
+
+  def markAsComplete(responses: ValidatorResponse): Unit = {
```
jonathonherbert: @jennygrahamjones this is something I'm a little confused by -- at the moment, `check` can be called _after_ `markAsComplete`, which before I added code in mitigation could cause e.g. the `'complete queued works'` spec to fail. But my Scala/Java is not good enough to understand why! I'd love to pair to understand why this might be happening.

jonathonherbert: e.g. in

```
val checkFuture = pool.check("test-1", "Example text")
validators.foreach(_.markAsComplete(responses))
```

`check` calls a validator for each category in turn, but `markAsComplete` is called (I think!) before `check` is done calling each validator.

app/services/ValidatorPool.scala
```diff
@@ -1,67 +1,136 @@
...
+
+  private def checkForCategory(id: String, text: String, categoryId: String): Future[ValidatorResponse] = {
+    val promise = Promise[ValidatorResponse]()
+    val future = promise.future
+    // Wrap our request in a lambda and add it to the queue, ready to be called.
+    val validate: () => Future[ValidatorResponse] = () => validators.get(categoryId) match {
+      case Some((_, validator)) =>
+        val eventualResult = validator.check(ValidatorRequest(categoryId, text))
+        eventualResult.andThen {
+          case Success(result) => {
```
jonathonherbert: @akash1810 this is where we maintain a reference to our caller as we're sat in the queue -- and where the code gets a little windy!

app/services/LanguageToolFactory.scala
```diff
@@ -53,10 +52,12 @@ class LanguageToolFactory(
...
 
 class LanguageTool(category: String, instance: JLanguageTool)(implicit ec: ExecutionContext) extends Validator {
+  def getId = "language-tool"
+
   def getCategory = category
 
-  def check(request: ValidatorRequest): List[RuleMatch] = {
-    instance.check(request.text).asScala.map(RuleMatch.fromLT).toList
+  def check(request: ValidatorRequest): Future[List[RuleMatch]] = {
+    Future.successful(instance.check(request.text).asScala.map(RuleMatch.fromLT).toList)
```
mbarton: I think this will execute `instance.check` synchronously on the current thread vs `Future { instance.check(...) }` which will be asynchronous in the `ExecutionContext`.

Not sure which one we want here, just flagging in case

app/services/ValidatorPool.scala
```diff
@@ -1,67 +1,121 @@
...
-}
+class ValidatorPool(val maxCurrentJobs: Int = 4, val maxQueuedJobs: Int = 100)(implicit ec: ExecutionContext, implicit val mat: Materializer) {
+  type TValidateCallback = () => Future[ValidatorResponse]
 
-class ValidatorPool(factory: ValidatorFactory)(implicit ec: ExecutionContext) {
+  case class ValidationJob(id: String, categoryId: String, text: String, promise: Promise[ValidatorResponse])
 
-  private val validators = new ConcurrentHashMap[String, Promise[Validator]]().asScala
+  private val validators = new ConcurrentHashMap[String, (Category, Validator)]().asScala
+  private val flow = Source.queue[ValidationJob](maxQueuedJobs, OverflowStrategy.dropNew)
```
mbarton: Definitely not against using Akka Streams here, the code is very clear. Just wondering what more it offers compared to use a Thread Pool and then native Java concurrent `BoundedQueue.offer` [methods](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ArrayBlockingQueue.html#offer-E-)?

test/scala/services/ValidatorPoolTest.scala
```diff
@@ -0,0 +1,148 @@
...
+    val validators = getValidators(1)
+    val pool = getPool(validators)
+    val eventuallyResult = pool.check("test-1", "Example text")
+    validators.head.markAsComplete(responses)
+    eventuallyResult.map { result =>
+      result shouldBe responses
+    }
+  }
+
+  "check" should "reject work that exceeds its buffer size" in {
```
mbarton: 👍 so good to have a test for this

app/services/ValidatorPool.scala
jonathonherbert: As discussed IRL -- we get backpressure and concurrency limits, but there's nothing to stop us implementing those things via a queue or similar. I'm inclined to use an abstraction that provides those things as part of its model rather than roll our own thing, but if there are good reasons to avoid Akka then I'm all 👂!

app/services/LanguageToolFactory.scala
jonathonherbert: Good question, and I'm not sure either -- let's chat 👍

jonathonherbert: You were correct, and stream concurrency was not a thing as a result. It's now resolved 👍

PR #14, opened 2019-11-11
Support multiple suggestion types

jonathonherbert:
There are lots of suggestion types we'd like to include in future, including the Wiki-like person lookups indicated in #12. This PR adds a more sophisticated `Suggestion` type, with a `type` field, so client consumers can disambiguate suggestions and handle them accordingly.

### DIFF
--- a/app/model/Suggestion.scala
+++ b/app/model/Suggestion.scala
@@ -1,0 +1,30 @@
+package model
+
+import play.api.libs.json.{Json, Reads, Writes}
+
+object Suggestion {
+  implicit val writes: Writes[Suggestion] = {
+    case textSuggestion: TextSuggestion =>
+      TextSuggestion.writes.writes(textSuggestion)
+  }
+}
+
+sealed trait Suggestion {
+  val `type`: String
+  val text: String
+}
+
+object TextSuggestion {
+  implicit val reads: Reads[TextSuggestion] = Json.reads[TextSuggestion]
+  implicit val writes = new Writes[TextSuggestion] {
+    def writes(suggestion: TextSuggestion) = Json.obj(
+      "type" -> suggestion.`type`,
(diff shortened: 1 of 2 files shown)


app/model/Suggestion.scala
```diff
@@ -0,0 +1,30 @@
...
+
+sealed trait Suggestion {
+  val `type`: String
+  val text: String
+}
+
+object TextSuggestion {
+  implicit val reads: Reads[TextSuggestion] = Json.reads[TextSuggestion]
+  implicit val writes = new Writes[TextSuggestion] {
+    def writes(suggestion: TextSuggestion) = Json.obj(
```
akash1810: Why do we need a custom writer? Could we use a [formatter](https://www.playframework.com/documentation/2.7.x/ScalaJsonCombinators#Format) instead?

jonathonherbert: IIRC if we don't create our own, the JSON is missing a `type` parameter. But perhaps there's a more elegant way to do this?

jonathonherbert: To cut down on duplication, for example, we could write ```
def writes(suggestion: TextSuggestion) = Json.obj("type" -> suggestion.`type`) ++ Json.writes[TextSuggestion](suggestion)
```

mbarton: Libraries like `play-json-extensions` provide [formatSealed](https://github.com/xdotai/play-json-extensions#formatsealed-uses-orelse-of-subclass-reads-in-random-order-careful-in-case-of-ambiguities-of-field-class-correspondances) to auto-generate serialisers and deserialisers for sealed traits but I think it's also fine just to manually define.

Probably best to define the Read and Writes on the companion object for the trait though to get exhaustivity checking

PR #15, opened 2019-11-11
Add the ability to validate entire documents at once

jonathonherbert:
_Dear reader -- thank you for your interest in Typerighter! Before you pay attention here, though, you should take a look at #14, which is upstream from this PR. 🙏 🙏_ 

This PR allows developers to map document division strategies to validators as they're added to the validator pool. This lets us send entire documents over the wire, and allows Typerighter to decide how to divide that work up.

This is useful because in addition to block-level work, which is what Typerighter currently does (one request per block), it's very likely we will want to support validators that need to reason about the entire document at once (we already have use-cases -- see e.g. #12). We may also want to support different ways of dividing up work to improve the user experience.

### Implementation details

`CheckStrategy` is the function type that maps a list of `Blocks` and `CategoryIds` into a list of jobs containing `Blocks` and `CategoryIds`. Each job - a `PartialValidationJob` - represent piece of work that can be tackled in isolation by the `ValidatorPool`.

The `validatorPool` maintains its old, `Future`-based interface, but in order to incrementally update the caller, I've introduced an event bus and the ability to subscribe and unsubscribe. This will be important later when we want to send the entire document over the wire at once, and receive incremental updates via e.g. a websocket connection.

Re: the event bus, I would be really interested to hear if anybody has any ideas about alternative approaches -- perhaps passing back a stream that's closed when work for that particular job is complete?

Finally, because the API caller needs to understand what particular piece of work has been completed when it receives an incremental response, the API response has been updated to let the caller know which blocks and category ids have been validated against in that response.


app/AppComponents.scala
```diff
@@ -67,4 +58,18 @@ class AppComponents(context: Context, identity: AppIdentity)
     rulesController,
     apiController
   )
+
+  /**
+    * Set up validators and add them to the validator pool as the app starts.
+    */
+  def initialiseValidators = {
```
jonathonherbert: Just a bit of a tidy 😁

app/services/ValidatorPool.scala
```diff
@@ -5,27 +5,52 @@ import java.util.concurrent.ConcurrentHashMap
...
+  *  - notify the job creator when that work is done.
+  */
+case class ValidationJob(blocks: List[TextBlock], categoryIds: List[String], requestId: String, promise: Promise[List[RuleMatch]], jobsInValidationSet: Integer)
+
+object ValidatorPool {
+  type CategoryIds = List[String]
+  type CheckStrategy = (List[TextBlock], CategoryIds) => List[PartialValidationJob]
+
+  def blockLevelCheckStrategy: CheckStrategy = (blocks, categoryIds) => {
+    for {
```
mbarton: nitpick but could this just be a `map` call rather than a for comprehension?

mbarton:
Looks sensible to me, I'm happy with the event bus as the intermediary.

> perhaps passing back a stream that's closed when work for that particular job is complete

This is also a decent use case for actors: spinning one up to receive partial responses and send them back to the client.

If you find different validators have radically different performance characteristics (eg the Stanford NLP is very CPU intensive) then you may want to bulkhead them onto separate EC2 instances (or Lambda functions if they are rarely invoked) to keep responsive to the user. We might want to upgrade the event bus to a full blown SQS queue in AWS or something.

PR #16, opened 2019-11-11
Rename `Validator` to `Matcher`

jonathonherbert:
...because it better expresses what we're doing! This is downstream from #14 and #15 -- please take a look at those first 😁


jonathonherbert:
Ah, bugger, thanks for spotting -- checked them in in a previous PR. Now removed.

PR #17, opened 2019-11-22
Add dedicated regex matcher

jonathonherbert:
LanguageTool is great, but its opinions about its rule model can be a problem. For example, it has an idea of 'suggestions', but it's difficult to add them programmatically (via xml in the rule description), and we can't distinguish between a suggestion and something we're definitely sure we can replace -- a `replacement` -- useful when considering e.g. an 'autofix' button for rules we're very sure about.

To this end, this PR introduces a matcher dedicated to regular expressions, with a separate rule model. Rules now extend a `BaseRule`, and so matchers can have their own opinions about the rules they're modelling.


app/model/BaseRule.scala
```diff
@@ -0,0 +1,24 @@
...
+trait BaseRule {
+  val id: String
+  val category: Category
+  val description: String
+  val suggestions: List[Suggestion]
+  val autoApplyFirstSuggestion: Boolean
+}
+
+object BaseRule {
+  def toJson(rule: BaseRule): JsObject = Json.obj(
```
mbarton: Could this be a `Writes[BaseRule]`? I think it would save writing out the formatter manually in `RuleMatch` below.

app/services/RegexMatcher.scala
```diff
@@ -0,0 +1,39 @@
...
+    Future.successful(rules.flatMap { checkRule(request, _) })
+  }
+
+  override def getRules(): List[RegexRule] = rules
+
+  override def getCategory(): String = category
+
+  private def checkRule(request: MatcherRequest, rule: RegexRule): List[RuleMatch] = {
+    request.blocks.flatMap { block =>
+      val iterator = rule.regex.findAllMatchIn(block.text)
```
mbarton: I think you can import `scala.collection.JavaConverters._` and convert using `.asScala` to call `.map` rather than manually running the iterator in a while loop

app/AppComponents.scala
```diff
@@ -62,14 +64,14 @@ class AppComponents(context: Context, identity: AppIdentity)
...
-  def initialiseMatchers = {
+  def initialiseMatchers: Future[Unit] = {
     for {
-      (rules, _) <- ruleResource.fetchRulesByCategory()
+      (rulesByCategory, _) <- ruleResource.fetchRulesByCategory()
     } yield {
-      rules.foreach { case (category, rules) => {
-        val (matcher, _) = languageToolFactory.createInstance(category.name, MatcherConfig(rules))
+      rulesByCategory.foreach { case (category, rules) => {
+        val matcher = new RegexMatcher(category.name, rules)
```
mbarton: Just checking my understanding here: of the box we get regex matchers for all the entries in the google sheet?

app/services/RegexMatcher.scala
jonathonherbert: Oh, that's _exactly_ what I was looking for. Thanks!

app/AppComponents.scala
jonathonherbert: For each category. So, not a matcher per rule.

app/model/BaseRule.scala
jonathonherbert: BaseRule is a trait, so I get `No apply function found for model.BaseRule`, but if this were possible it'd be ideal -- I'll ask about.

mbarton: Ah yeah, I think you'd have to write all the fields out manually which wouldn't be much fun. No problem, it's fine as it is

PR #18, opened 2019-11-26
Small tweak to key location

sihil:
This was left in my local tree so presumably is a good idea?


jonathonherbert:
This has since been superceded by `/account/services/logging.stream` in a [subsequent PR](https://github.com/guardian/typerighter/pull/21).

PR #20, opened 2019-11-27
Streaming API endpoint

sihil:
Add a new `checkStream` endpoint that is identical to `check`, but gives a chunked response.

The reponse contains multiple JSON documents separated with a `RECORD_SEPARATOR` character per the [json-seq spec.](https://en.wikipedia.org/wiki/JSON_streaming#Record_separator-delimited_JSON_2)

There are two advantages to this approach:

- It allows Typerighter to reason about the whole document in one go, opening the door to matchers that care about the entire doc.
- It's less work for the client by a factor of about two. Here's the total browser processor time in Chrome, for the default long read in the prosemirror-client, averaged over 3 attempts on a macbook pro 2019 in ms:

|Before|After|
|-|-|
|1614|784|

## Why not server-sent events, or websockets?

We explored both of these:
- Server sent events would be perfect but for the fact that the current Javascript API [(spec)](https://html.spec.whatwg.org/multipage/server-sent-events.html) does not allow you to specify a verb in the request, or provide a request body. This limits us to query params when sending text, which across browsers is about [2000 characters](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers) – unacceptably small for entire documents.
- Websockets fit the use case nicely, but the bidirectional communication once a connection is established is a larger surface area than we need. Sending an initial payload including the document, and streaming back a response that was considered complete at its end, was a better semantic fit. In addition to this, the integration Play provides with streams for chunked responses means we need to write less code – the websocket approach required an Actor-based implementation to handle incoming messages.

## Dev notes

I've confirmed in Wireshark that the chunks in the TCP correspond to the records in the stream. However, there's no guarantee that those chunks will be what browsers provide via their APIs – see the [sibling PR](https://github.com/guardian/prosemirror-typerighter/pull/214) in prosemirror-typerighter for details.

## How to test

The automated tests should pass, and cover the core stream behaviour, including successfully issuing responses, failures happening within the `matcherPool` stream as a result of backpressure, and failures caused by the stream created by `checkStream` specifically.

Spin up the prosemirror-typerighter client on [this branch](https://github.com/guardian/prosemirror-typerighter/tree/jsh/default-to-ws), which adds an adapter that handles a chunked response, and run a check against the default document. You should be able to see that a single request goes over the wire, and that responses stream back into the document over time. This change is a no-op – everything should work as it did.


app/controllers/ApiController.scala
```diff
@@ -19,12 +20,7 @@ class ApiController(
       case Right(check) =>
         matcherPool
           .check(check)
-          .map { matches =>
-            val response = MatcherResponse(
-              matches = matches,
-              blocks = check.blocks,
-              categoryIds = check.categoryIds.getOrElse(matcherPool.getCurrentCategories.map { _._1 })
-            )
+          .map { response =>
```
jonathonherbert: So much cleaner 👍 👍

app/services/MatcherPool.scala
```diff
@@ -49,14 +53,12 @@ object MatcherPool {
...
-class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, val checkStrategy: MatcherPool.CheckStrategy = MatcherPool.documentPerCategoryCheckStrategy)(implicit ec: ExecutionContext, implicit val mat: Materializer) {
-  type JobProgressMap = Map[String, Int]
+class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, val defaultCheckStrategy: MatcherPool.CheckStrategy = MatcherPool.documentPerCategoryCheckStrategy)(implicit ec: ExecutionContext, implicit val mat: Materializer) {
 
   private val matchers = new ConcurrentHashMap[String, (Category, Matcher)]().asScala
-  private val eventBus = new MatcherPoolEventBus()
   private val queue = Source.queue[MatcherJob](maxQueuedJobs, OverflowStrategy.dropNew)
     .mapAsyncUnordered(maxCurrentJobs)(runValidationJob)
-    .to(Sink.fold(Map[String, Int]())(markJobAsComplete))
+    .to(Sink.ignore) // throw away result
```
jonathonherbert: Nit: arguable whether this comment is necessary!

app/services/RegexMatcher.scala
```diff
@@ -1,15 +1,21 @@
...
+import scala.concurrent.{ExecutionContext, Future}
 
 class RegexMatcher(category: String, rules: List[RegexRule]) extends Matcher {
   def getId() = "regex-validator"
 
-  override def check(request: MatcherRequest): Future[List[RuleMatch]] = {
-    Future.successful(rules.flatMap { checkRule(request, _) })
+  override def check(request: MatcherRequest)(implicit ec: ExecutionContext): Future[List[RuleMatch]] = {
+    Future {
+      Logger.info(s"Running regex matcher ${request.blocks.map(_.id).mkString(",")} on thread ${Thread.currentThread().getName}")
```
jonathonherbert: Is this log message still necessary? We're quite verbose in the `MatcherPool` already.

sihil: Probably not.

sihil:
I hope to pick this up when WFH on Friday. I did try to write a test for checkStream but the MockMatcher didn't lend itself to doing so without significant changes. Will have another go though.

app/services/MatcherPool.scala
```diff
@@ -75,24 +77,44 @@ class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, va
...
+    }
+
+    Logger.info(s"Validation job with id: ${query.requestId} received. Checking categories: ${categoryIds.mkString(", ")}")
+
+    val jobs = createJobsFromPartialJobs(checkStrategy(query.blocks, categoryIds), query.requestId)
+    Logger.info(s"Created ${jobs.size} jobs")
+    val eventualResponses = jobs map offerJobToQueue
+    val responsesAsSources = eventualResponses map Source.fromFuture
+    val mergedResponses = responsesAsSources.foldLeft(Source.empty[MatcherJobResult]) { case (source, acc) =>
+        acc.merge(source)
```
jonathonherbert: Just FYI, I took a look and we can also create streams directly from lists of `Futures`, which makes this a one liner --

```
val responseStream = Source(eventualResponses).mapAsyncUnordered(1)(identity)
```

-- I can't see anything funny in the `mapAsyncUnordered` call re: performance or parallelism (there's no blocking work to do) but you may spot a reason this is unsuitable!

jonathonherbert:
cc @sihil – this is now ready for review. No unit tests due to auth – how do we mock a trait already inherited by a controller? still looking  – in the mean time, this is testable in a local environment by amending the prosemirror-typerighter client to point to the websocket environment ([example branch](https://github.com/guardian/prosemirror-typerighter/tree/jsh/default-to-ws))

sihil:
Still want this @jonathonherbert. Am closing all old PRs in my name as part of some spring cleaning (is it spring yet???(

jonathonherbert:
It's always spring in my heart. Don't close this, we'll want it, soon-ish 😁

sihil:
🌸 
Do you want to push it as a new PR with you as an owner?

jonathonherbert:
No opinion on my side but happy to do so, esp. if you've got notification noise as a result of ownership.

rhystmills (approved):
Running as expected for me with a [suitable consuming project](https://github.com/guardian/prosemirror-typerighter/pull/214#pullrequestreview-1066300534). Code looks good and tests passing.

PR #21, opened 2019-12-08
Move typerighter onto internet by adding panda verification

sihil:
At present Typerighter is only available in the office building which doesn't match the availability of Composer that is available on the internet at large. By moving Typerighter to the internet this will make the service available regardless of where the user is situated.

This PR is untested but should be fairly close to migrating typerighter onto the internet instead of intranet.

We have:
 - included the panda verification library
 - hardcoded the configuration for now
 - updated the CFN template to use a external LB with separate subnets to the instances

Note that this will never issue a cookie, we assume the calling service (e.g. Composer) will do that. We provide a one hour grace period so that the panda cookie can be out of date.

Now tested in CODE.

Thoughts
-----------
This currently forces all users to use pan domain authentication which is orthoganal to our aims to make this a project useful to everyone. Perhaps we can have a switch for this that defaults to off?

 - [ ] Add configuration options
 - [ ] Fallback to no auth if config is not set? (is this dangerous?)

Another completely different option is to introduce an API gateway that does the auth and not touch the code in the project at all.

Rollout
-------
 - [x] Add permission to access S3 bucket
 - [ ] Rollout internally so internal users are authenticated
 - [ ] Reduce TTL of DNS entries
 - [ ] Switch to internet LB
 - [ ] Update DNS

Note that this will interupt service for a short period governed by the speed of change and the TTL of the DNS entries


jonathonherbert:
Great work -- I think definitely necessary for now. There's nothing stopping us from testing on a CODE environment, no?

PR #24, opened 2020-01-14
Add basic onboarding documentation

aug24:


conf/routes
```diff
@@ -8,9 +8,9 @@ GET     /assets/*file               controllers.Assets.versioned(path="/public",
 GET     /                           controllers.HomeController.index
 GET     /healthcheck                controllers.HomeController.healthcheck
 
-POST    /rules                      controllers.RulesController.refresh
+POST    /refresh                    controllers.RulesController.refresh
```
jonathonherbert: This needs to be `/rules`, or our refresh button will break.

docs/01-running-locally.md
```diff
@@ -0,0 +1,48 @@
+# Basic developer run
+
+`sbt run` and visit `localhost:9000` to confirm the service came up correctly.
+
+## Demonstration of null request
+
+Each request has the following format, of zero or more text blocks, and zero or more rule sets to check.
+
+```
```
jonathonherbert: If `jq` is a dependency of these commands, we should specify it above -- but I'd be tempted to remove it for now, on the assumption that devs who want to use it, will.

conf/routes
jonathonherbert: ... just checked the code, and I stand corrected 😁

docs/01-running-locally.md
jonathonherbert: (super minor)

PR #26, opened 2020-01-16
Upversion Play to 2.8

aug24:


app/services/LanguageToolFactory.scala
```diff
@@ -8,14 +8,14 @@ import org.languagetool.rules.spelling.morfologik.suggestions_ordering.Suggestio
 
 import collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
-import play.api.Logger
+import play.api.Logging
```
jonathonherbert: _slaps forehead_ of course! ... 😁

PR #30, opened 2020-01-17
Use Set not List for comparison to avoid stochastic effects

aug24:
With a list, stochastic effects of queuing/polling mean that the eventual order is not consistent.  This makes testing inconsistent.

As the ordering is actually irrelevent, switching to an unordered collection (Set) removes this problem.

### DIFF
--- a/script/citest
+++ b/script/citest
@@ -1,3 +1,3 @@
-#!/bin/sh
+#!/bin/sh -e
 
 sbt clean compile test riffRaffNotifyTeamcity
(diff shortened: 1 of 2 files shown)


script/citest
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
akash1810: can we add a `-e`?

aug24: Oh, OK then.

PR #35, opened 2020-02-27
Add an nginx mapping file

jonathonherbert:
... which makes things _official_.


jonathonherbert:
@akash1810 Thanks! Odd, looks like we have a flakey test, I'll take a look.

PR #36, opened 2020-02-27
Add a CAPI proxy

jonathonherbert:
Add a set of endpoints to proxy CAPI searches, in preparation for the management frontend's rule audit service.

This service will allow users to query CAPI for content, and then run this content through our matcher service for audit.


app/services/ContentClient.scala
```diff
@@ -0,0 +1,42 @@
...
+import play.api.libs.json.{Json, Reads, Writes}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object CapiSearchParams {
+  implicit val writes: Writes[CapiSearchParams] = Json.writes[CapiSearchParams]
+  implicit val reads: Reads[CapiSearchParams] = Json.reads[CapiSearchParams]
+}
+
+case class CapiSearchParams(query: String, tags: Option[List[String]] = None, sections: Option[List[String]] = None)
```
akash1810: I think the types can be simpler here. `Option[List[String]]` could be a `List[String]` with a default value of `List.empty`.

jonathonherbert: True. This is the result of me servicify-ing controller code – IIRC using `List.empty` results in a non-optional field when deserialising from JSON, so we'll need a new type in the controller, but that's fine 👍

akash1810: Yeah, making the controller a translator from request and all it's properties into strictly typed cases classes makes this easier to reason about and test IMO.

app/controllers/CapiProxyController.scala
```diff
@@ -0,0 +1,40 @@
...
+          query.query,
+          query.tags.getOrElse(Nil),
+          query.sections.getOrElse(Nil))
+        .map { result => Ok(result.asJson.toString).as(JSON) }
+      case Left(error) => Future.successful(BadRequest(s"Invalid request: $error"))
+    }
+  }
+
+  def searchTags(queryStr: String) = Action.async {
+    contentClient.searchTags(queryStr).map { result => Ok(result.asJson.toString).as(JSON) }
```
akash1810: `Ok(result.asJson.toString).as(JSON)` I think we talked about this IRL, it still confuses me though 🤔

jonathonherbert: I had to remind myself! It's circe JSON, so Play doesn't know what to do with it – we need to pass a string and add the content type manually.

We might be able to come up with some sugar for this if we end up writing it a lot.

jonathonherbert:
@akash1810 – I sorted a one line merge conflict and now require reapproval – if you could be so kind 😁

PR #37, opened 2020-02-28
Add a vision document

jonathonherbert:
Adds a vision document based on discussions with @sihil, Stephan Fowler and Max Walker. Its structure is based on [the Grid's](https://github.com/guardian/grid) vision document.

This is the first draft of this living document. Suggestions welcome, and we should expect it to change over time.

### DIFF
--- a/vision.md
+++ b/vision.md
@@ -1,0 +1,70 @@
+# Vision
+
+The goal of Typerighter is to provide a [universally available](#universal) document checking service, via  an [easy-to-use](#easy-to-use) and [minimal interface](#minimal), that is [accurate](#accurate), [transparent](#transparent), and [responsive to change](#responsive-to-change).
+
+## Universal
+
+Good proofing tools should provide value to users wherever they write.
+
+Typerighter should be available for use across our internal toolset and beyond it. It should be easy for other applications to integrate with it, regardless of platform or publishing cadence. It should just care about text.
+
+### KPIs
+
+- Adoption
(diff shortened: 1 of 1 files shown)


vision.md
```diff
@@ -0,0 +1,70 @@
+# Vision
+
+The goal of Typerighter is to provide a [universally available](#universal) document checking service, via  an [easy-to-use](#easy-to-use) and [minimal interface](#minimal), that is [accurate](#accurate), [transparent](#transparent), and [responsive to change](#responsive-to-change).
+
+## Universal
+
+Good proofing tools should provide value to users wherever they write.
+
+Typerighter should be available for use across our internal toolset and beyond it. It should be easy for other applications to integrate with it, regardless of platform or publishing cadence. It should just care about text.
```
sihil: A further paragraph here being a bit more explicit about availability. Firstly by providing a non-exhaustive list of example places we'd expect to have it. Secondly a principle such as "it should be available to check any field that is ultimately published and readable by the public".

vision.md
```diff
@@ -0,0 +1,70 @@
...
+
+### KPIs
+
+- Adoption
+- Number of tools using the service
```
sihil: Coverage of each tool? e.g. A tool might use it in one field but not everywhere...

jonathonherbert:
Thanks for the feedback @sihil. Agreed on the tense, and I'll go through the comments case by case.

PR #41, opened 2020-04-08
Add a skeleton for the single-page audit app

jonathonherbert:
## What does this PR do?

It adds a skeleton for the single-page audit app.

## What the devil is the single-page audit app?

It's an app to help rule makers understand the impact of their rules. The hypothetical flow looks something like –

- Rule-maker creates rule, would like to see how it performs in production
- Rule-maker uses app to scan CAPI for content that matches intended use of rule, e.g. sport for footballer names, some tag for common legal problems etc.
-Rule-maker uses app to run parts of the searched content through Typerighter, and uses the UI to audit the matches that are found.

In this way, it should be clearer to the rule-maker how rules are likely to perform in production.

## What does success look like?

Fire up the app in TEST. You should see a new `audit` tab. If the text 'Audit app' appears, the JS is working, as this text comes from the React app.

Screenshot 2020-04-02 at 18 34 15

## Implementation details

This is served over a hot-reloadable webpack server in `dev`, and build into a bundle served with Play's assets and injected into the body of a new `audit` page.

In the future, it'd probably be better to incorporate the current Typerighter Play views into the SPA and drive it from the API, rather than have a mix of Play templates and React apps, but this is a quick way of shipping a rich client for now.

### DIFF
--- a/app/controllers/CapiProxyController.scala
+++ b/app/controllers/CapiProxyController.scala
@@ -2,32 +2,20 @@ package controllers
 
 import com.gu.contentapi.json.CirceEncoders._
 import io.circe.syntax._
-import play.api.libs.json.JsValue
 import play.api.mvc._
-import play.api.libs.json.{Json, Reads, Writes}
 import services.{ContentClient}
 
 import scala.concurrent.{ExecutionContext, Future}
 
-object CapiSearchParams {
-  implicit val writes: Writes[CapiSearchParams] = Json.writes[CapiSearchParams]
-  implicit val reads: Reads[CapiSearchParams] = Json.reads[CapiSearchParams]
-}
-
-case class CapiSearchParams(query: String, tags: Option[List[String]] = None, sections: Option[List[String]] = None)
-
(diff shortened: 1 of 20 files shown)


app/controllers/CapiProxyController.scala
```diff
@@ -2,32 +2,20 @@ package controllers
...
-  def searchContent: Action[JsValue] = Action.async(parse.json) { request =>
-    request.body.validate[CapiSearchParams].asEither match {
-      case Right(query) => contentClient.searchContent(
-          query.query,
-          query.tags.getOrElse(Nil),
-          query.sections.getOrElse(Nil))
-        .map { result => Ok(result.asJson.toString).as(JSON) }
-      case Left(error) => Future.successful(BadRequest(s"Invalid request: $error"))
-    }
+  def searchContent(query: String, tags: Option[List[String]], sections: Option[List[String]]) = Action.async {
```
jonathonherbert, line 13: @aug24 This change moves the search parameters for the CAPI client from the request body to url params; Play will do the parsing and typechecking for us, and it's a better fit for what should be a GET request.

This will be useful in upcoming work, but there's no particular reason to include the change in this PR, so if we'd like it moved out then please shout 👍

PR #43, opened 2020-04-08
Add a searchable feed to the CAPI audit client

jonathonherbert:
This PR adds searchable feed to the CAPI audit client –

capi-search

### This looks an awful lot like the [Fronts](https://github.com/guardian/facia-tool) client search! Why not use that?

An excellent question! We could use that, but the code is quite complex. This is a simpler implementation that gets us off the ground. It's also more amenable to other improvements – for example, using the searchable `Chips` component that's currently available in [Threads](https://github.com/guardian/threads) – that I hope will eventually prove to be useful in the Fronts project.

### What's with `createAsyncResourceBundle`? Haven't I seen something similar before?

You have! It's also used in the Fronts project. I've taken the opportunity to package it as a library for use in this app, `redux-bundle-creator` – it's not very well documented (yet!), and the structure is still quite new, but it's well tested, and the typesafety of the package is significantly better than the code in Fronts.

I'd like to improve the documentation and backport this project into Fronts if possible, increasing typesafety in that project and removing some LoC.


rule-audit-client/src/components/capiFeed/CapiResults.tsx
```diff
@@ -0,0 +1,27 @@
...
+type IProps = ReturnType<typeof mapStateToProps>;
+
+const CapiResults = ({ content }: IProps) => (
+  <div className="mt-2">
+    {Object.values(content).map(article => (
+      <div className="card mt-2">
+        <div className="card-body">
+          <div className="card-text">{article.webTitle}</div>
+          <div className="badge badge-warning mr-1">2</div>
+          <div className="badge badge-danger">4</div>
```
akash1810, line 16: What does `4` mean here? (Same for `2` above).

rule-audit-client/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+const urls = {
+    capiQueryUrl: 'capi'
```
akash1810, line 2: How does `'capi'` get used? Should it be a URL?

rule-audit-client/src/services/Capi.ts
```diff
@@ -0,0 +1,77 @@
...
+  results: TContent[] | undefined;
+};
+
+export type CapiContentResponse = CapiResponse<CapiContent>;
+
+export type CapiTagsResponse = CapiResponse<CapiTag>;
+
+export type CapiSectionsResponse = CapiResponse<CapiSection>;
+
+export type CapiContent = {
```
akash1810, line 19: IIRC @alexduf was working on adding TS types to CAPI, maybe we can use them here?

rule-audit-client/src/constants.ts
jonathonherbert: This would be better named as `capiQueryPath` – it's the path to query the capi controller, mirrored in `conf/routes` in the main project.

It'd be nice to derive both the paths and the types from the play project, wouldn't it? 😁

rule-audit-client/src/components/capiFeed/CapiResults.tsx
jonathonherbert: It's a placeholder for different levels of typerighter warning – they don't mean anything at the moment! They're used in an upcoming PR that adds actual validation ... but I appreciate they don't mean much now.

rule-audit-client/src/services/Capi.ts
jonathonherbert: Absolutely! I don't know what the status of this project is, but I'm very keen to use them as soon as they're available.

PR #44, opened 2020-06-01
Fetch typerighter matches for the current search

jonathonherbert:
## What does this change?

Adds the ability to fetch typerighter matches for the current search.

fetch-matches

Spin Typerighter up locally, or in CODE, and open the rule audit client. On clicking 'Find matches for this search', you should see the matches start to roll in 😁


rule-audit-client/src/services/typerighter.ts
```diff
@@ -0,0 +1,22 @@
...
+  articleId: string,
+  blocks: IBlock[]
+): Promise<IMatcherResponse> => {
+  const response = await fetch(urls.matches, {
+    method: "POST",
+    headers: new Headers({
+      "Content-Type": "application/json"
+    }),
+    body: JSON.stringify({
+      requestId: `audit-${articleId}`,
```
mbarton, line 16: Should this be unique across multiple requests against the same article? I'd expect `requestId` to be unique across all the logs if I was searching in ELK

rule-audit-client/src/utils/prosemirror.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const appSchema = new Schema({
+  nodes: addListNodes(schema.spec.nodes as any, "paragraph block*", "block"),
+  marks
+});
+
+const parser = DOMParser.fromSchema(appSchema);
+
+export const getBlocksFromHtmlString = (htmlStr: string) => {
+  const element = document.createElement("div");
```
mbarton, line 16: I don't think it matters here since we trust the source of our content but I've used [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) in the past to parse HTML

rule-audit-client/src/services/typerighter.ts
jonathonherbert: Yes, this is a bit lazy! Quite right, I'll add a uuid.

rule-audit-client/src/utils/prosemirror.ts
jonathonherbert: This is absolutely what should be happening here, good catch 👍

rule-audit-client/src/services/typerighter.ts
sihil: In terms of logging the method we use elsewhere is an `X-GU-LoggingContext` header. See https://github.com/guardian/flexible-content/blob/master/flexible-content-common/src/main/scala/com/gu/flexiblecontent/util/LoggingContext.scala and https://github.com/guardian/flexible-content/blob/master/composer/src/js/app.js#L393-L419

You could include `contentId` there for example. `sessionId` would be cute too if you can access it.

jonathonherbert: Ah, nice – I'll keep this the signature for now, but I've raised https://github.com/guardian/typerighter/issues/46 to capture.

rule-audit-client/src/utils/prosemirror.ts
jonathonherbert: Hm, on further consideration I don't think `DocumentFragment` helps us here – you must still append a valid Node to the created fragment. As you've approved, I'll merge for now, although happy to discuss and amend if I've misunderstood!

On my travels to understand this better, though, I did discover DOMParser, which has an advantage in that it doesn't trigger browser fetches on parse for elements with src attributes, so I've used that instead.

PR #49, opened 2020-06-25
Add CAPI chip search to replace select2

jonathonherbert:
## What does this change?

Adds a chip input for the CAPI search to replace the current select2 implementation.

chips

This is arguably more concise than having <n> selects for <n> possible fields, as is the case in the CAPI search of e.g. the current Fronts tool. I'd definitely consider this as a PoC for improving CAPI searches in other tools.

Fire up Typerighter and navigate to the rule audit client. You should be able to use the chip search to search via string, or by adding tags and sections.

## Known issues

- At the moment, the input can easily become full up if more than a few chips are added. We can solve in the short term by widening the input, and in the longer term by improving the chip component to more gracefully handling wrapped content.
- The theming at the moment consists of overrides to css variables, which is less than ideal, and something that I'd love to improve by [providing an API in Threads](https://github.com/guardian/threads/issues/4).

## Notes

This is now possible due to recent changes in the [Threads](https://github.com/guardian/threads/pull/3) package that allow us to run asynchronous searches in response to changes to the chip input – thanks @itsibitzi for your help with this work! This change was necessary to dynamically fetch tags and sections – cf. e.g. [Async Select](https://react-select.com/async).


rule-audit-client/src/components/capiFeed/CapiResults.tsx
```diff
@@ -3,15 +3,15 @@ import AppTypes from "AppTypes";
...
-  selectors as uiSelectors,
-} from "redux/modules/ui";
+import { selectors as uiSelectors } from "redux/modules/ui";
 import { CapiContentWithMatches } from "services/capi";
 import { notEmpty } from "utils/predicates";
 import CapiFeedItem from "./CapiFeedItem";
 
 type IProps = ReturnType<typeof mapStateToProps>;
 
+const articleNumberFormat = new Intl.NumberFormat("en-GB");
```
akash1810, line 13: 😍

rule-audit-client/src/components/capiFeed/CapiSearchOptions.tsx
```diff
@@ -2,97 +2,110 @@ import * as React from "react";
...
+        })) || []
+      );
+    },
+  },
+];
+
+const getSearchEntitiesFromQueryElements = (elements: QueryElement[]) => ({
+  query: elements
+    .filter((element) => element.type === "text")
+    .reduce((acc, el) => `${acc} ${el.value}`, ""),
```
akash1810, line 55: why the empty string here?

rule-audit-client/src/services/capi.ts
```diff
@@ -67,8 +67,11 @@ export const fetchCapiSearch = async (
 ): Promise<CapiContentResponse> => {
   const params = new URLSearchParams();
   params.append("query", query);
-  tags.map(_ => params.append("tags", _));
-  sections.map(_ => params.append("sections", _));
+
+  // Do not include empty tag or section values.
+  tags.filter(_ => _).forEach(_ => params.append("tags", _));
```
akash1810, line 72: This filter looks interesting, I kinda like it! 👍 for the explaining comment.

rule-audit-client/src/components/capiFeed/CapiSearchOptions.tsx
jonathonherbert: As I understand it the accumulator starts with an empty string – otherwise we'll get a string 'undefined' as the value `undefined` is coerced.

Bit lazy as we'll always have two leading spaces at the start of this query, which I hadn't spotted. In fact, I'd completely forgotten about `.join`, which would be better! – I'll use that.

rule-audit-client/src/components/capiFeed/CapiResults.tsx
jonathonherbert: I didn't know about this until I tried to do it!

PR #50, opened 2020-07-06
Add a 'search for matches' mode to the rule audit client

jonathonherbert:
~Depends on #52 – please review that PR first for a smaller diff of the state management code.~ This is now ready for review.

## What does this change?

This PR adds a 'search for matches' mode, in addition to the standard CAPI search. This mode runs a CAPI search, finds Typerighter results for it, and repeats, until n articles with matches have been found. N is specified by the user.

search-matches

## Why is this useful?

At the moment, it's easy to answer the question, 'what results does Typerighter give for the results of this particular CAPI search'. But it's difficult to answer the question, 'how does Typerighter perform for this CAPI search more generally'?

By searching for articles with matches, and not just articles, we can get a better idea of how rules will be applied across the corpus more generally – both by finding lots of articles with matches, and also by giving us an idea of how often these matches will apply, to spot e.g. noisy rules.

In the rule audit client, click on 'search by matches', and then click the search button. The app should search CAPI and run articles through Typerighter until the number of articles with matches specified in the number input have been found.

### DIFF
--- a/rule-audit-client/src/components/capiFeed/CapiFeedItem.tsx
+++ b/rule-audit-client/src/components/capiFeed/CapiFeedItem.tsx
@@ -18,7 +18,6 @@ type IProps = ReturnType<typeof mapStateToProps> &
 
 const CapiFeedItem = ({
   doSelectArticle,
-  selectedArticle,
   article,
   isLoading,
 }: IProps) => {
--- a/rule-audit-client/src/redux/modules/searchMatches/__tests__/thunks.spec.ts
+++ b/rule-audit-client/src/redux/modules/searchMatches/__tests__/thunks.spec.ts
@@ -189,9 +189,9 @@ describe("doSearchMatches", () => {
 
     // We call CAPI, and it's returned two articles.
     // We then call Typerighter with those articles, and Typerighter
-    // returns matches for one. We call CAPI again, but the user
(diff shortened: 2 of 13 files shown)


rule-audit-client/src/components/capiFeed/CapiFeedItem.tsx
```diff
@@ -18,7 +18,6 @@ type IProps = ReturnType<typeof mapStateToProps> &
 
 const CapiFeedItem = ({
   doSelectArticle,
```
jonathonherbert, line 20: This was unused.

rule-audit-client/src/redux/modules/searchMatches/__tests__/thunks.spec.ts
```diff
@@ -189,9 +189,9 @@ describe("doSearchMatches", () => {
 
     // We call CAPI, and it's returned two articles.
     // We then call Typerighter with those articles, and Typerighter
-    // returns matches for one. We call CAPI again, but the user
-    // cancels the search before the second CAPI request is processed,
-    // so nothing else should go to Typerighter.
+    // returns no matches. Normally we'd continue to call CAPI,
```
jonathonherbert, line 192: Oops! This was intended to be in #52. Ported forward here – still relevant.

PR #52, opened 2020-07-13
Add a 'search matches' function to the state

jonathonherbert:
~Greetings, gentle reader! Please review #51 first – this PR depends upon it.~ This is now ready for review.

## What does this change?

This PR adds a 'search for matches' function. This function runs a CAPI search, finds Typerighter results for it, and repeats, until n articles with matches have been found. N is specified by the user.

It doesn't expose this functionality in the UI yet – a previous PR includes this but is very large and thus harder to review.

## Why will this eventually be useful?
At the moment, it's easy to answer the question, 'what results does Typerighter give for the results of this particular CAPI search'. But it's difficult to answer the question, 'how does Typerighter perform for this CAPI search more generally'?

By searching for articles with matches, and not just articles, we can get a better idea of how rules will be applied across the corpus more generally – both by finding lots of articles with matches, and also by giving us an idea of how often these matches will apply, to spot e.g. noisy rules.

The tests pass, and describe the behaviour we expect for the function.

### DIFF
--- a/rule-audit-client/jest.config.js
+++ b/rule-audit-client/jest.config.js
@@ -1,6 +1,8 @@
 module.exports = {
-  moduleFileExtensions: ['ts', 'tsx', 'js'],
+  moduleFileExtensions: ["ts", "tsx", "js"],
+  modulePaths: ["<rootDir>/src"],
+  testMatch: ["**/?(*.)+(spec|test).[jt]s?(x)"],
   transform: {
-    '^.+\\.(ts|tsx|js)$': 'ts-jest'
-  }
+    "^.+\\.(ts|tsx|js)$": "ts-jest",
+  },
 };
--- a/rule-audit-client/src/components/App.tsx
+++ b/rule-audit-client/src/components/App.tsx
@@ -2,9 +2,11 @@ import { hot } from "react-hot-loader/root";
 import * as React from "react";
 import { Provider } from "react-redux";
 
-import store from "redux/store";
+import { createStore } from "redux/store";
 import CapiFeed from "./capiFeed/CapiFeed";
 import Document from "./document/Document";
+
+const store = createStore();
 
 const App = () => (
   <Provider store={store}>
--- a/rule-audit-client/src/components/capiFeed/CapiResults.tsx
+++ b/rule-audit-client/src/components/capiFeed/CapiResults.tsx
@@ -11,32 +11,13 @@ import CapiFeedItem from "./CapiFeedItem";
 type IProps = ReturnType<typeof mapStateToProps>;
 
 const articleNumberFormat = new Intl.NumberFormat("en-GB");
+const checkboxId = "capi-results__show-all-articles";
 
-const filterArticles = (
-  articles: (CapiContentWithMatches | undefined)[],
-  showArticlesWithMatchesOnly: boolean
-) =>
-  showArticlesWithMatchesOnly
-    ? (articles.filter(
-        (_) => (_ !== undefined && !_.meta.matches) || _?.meta.matches.length
-      ) as CapiContentWithMatches[])
-    : (articles.filter((_) => _ !== undefined) as CapiContentWithMatches[]);
-
-const CapiResults = ({ articles, isLoading, pagination }: IProps) => {
-  const [
-    showArticlesWithMatchesOnly,
-    setShowArticlesWithMatchesOnly,
-  ] = useState(false);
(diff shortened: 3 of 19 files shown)


rule-audit-client/jest.config.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
-  moduleFileExtensions: ['ts', 'tsx', 'js'],
+  moduleFileExtensions: ["ts", "tsx", "js"],
```
jonathonherbert, line 2: Necessary to import fixtures, and ensure that imports relative to the 'src' directory are resolved correctly in tests.

rule-audit-client/src/components/App.tsx
```diff
@@ -2,10 +2,12 @@ import { hot } from "react-hot-loader/root";
 import * as React from "react";
 import { Provider } from "react-redux";
 
-import store from "redux/store";
+import { createStore } from "redux/store";
```
jonathonherbert, line 5: This change lets us use `createStore` multiple times in our tests.

rule-audit-client/src/components/capiFeed/CapiResults.tsx
```diff
@@ -11,32 +11,13 @@ import CapiFeedItem from "./CapiFeedItem";
...
-  const filteredArticles = filterArticles(
-    articleArray,
-    showArticlesWithMatchesOnly
-  );
+const CapiResults = ({ articleIds, isLoading, pagination }: IProps) => {
   const totalArticles = pagination
     ? articleNumberFormat.format(
-        filteredArticles.length < pagination.pageSize
-          ? filteredArticles.length
+        articleIds.length < pagination.pageSize
```
jonathonherbert, line 19: A slight simplification of this UI to remove some presentation – this anticipates another PR. As this functionality isn't yet used we're safe to do so 👍

rule-audit-client/src/redux/modules/capiContent/index.ts
```diff
@@ -1,82 +1,15 @@
...
-import { getBlocksFromHtmlString } from "utils/prosemirror";
-import { fetchTyperighterMatches } from "services/typerighter";
-import { notEmpty } from "utils/predicates";
+import { CapiContentWithMatches } from "services/capi";
+import * as _selectors from './selectors';
 
-const bundle = createAsyncResourceBundle<CapiContentWithMatches, {}, "capi">(
+export * as thunks from './thunks';
+
+export const { actions, reducer, ...bundle } = createAsyncResourceBundle<CapiContentWithMatches, {}, "capi">(
```
jonathonherbert, line 8: I've moved the thunks into their own file for convenience sake, as this file was getting a little large.

rule-audit-client/src/redux/modules/searchMatches/reducer.ts
```diff
@@ -0,0 +1,31 @@
...
+  searchMatchesArticleIds: [] as string[]
+};
+
+export type SearchMatchesState = typeof initialState
+
+const reducer = createReducer(initialState)
+  .handleAction(doSetSearchMatchesLimit, (state, action) => ({
+    ...state,
+    searchMatchesLimit: action.payload,
+  }))
```
jennygrahamjones, line 16: I don't think I've seen "typesafe-actions" `createReducer` used before. This is so much more concise and readable than the usual reducer structure! 👍

rule-audit-client/src/redux/modules/searchMatches/__tests__/thunks.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    await thunk;
+
+    const state = store.getState();
+
+    // We call CAPI, and it's returned two articles.
+    // We then call Typerighter with those articles, and Typerighter
+    // returns matches for one. We call CAPI again, but the user
+    // cancels the search before the second CAPI request is processed,
+    // so nothing else should go to Typerighter.
```
jennygrahamjones, line 194: Really appreciate the thorough explanation of these different scenarios 👍

jonathonherbert: Oh, that's great! I found it difficult to express the scenarios as code – glad it was helpful.

jonathonherbert: The only problem is that it's easy to write things that are out of sync with the code! I've just spotted this below – I'll amend.

PR #53, opened 2020-07-13
Update 01-running-locally.md

jennygrahamjones:
Adds a pointer towards the rules audit client README, for instances when you're testing something related to the client locally. I forget this is sometimes a required step, so thought it might be worth adding to the main docs on running locally.


jonathonherbert:
Great point @jennygrahamjones. Eventually, this client will server the whole of the frontend for the typerighter API and won't seem so tucked away 👍

PR #61, opened 2020-08-12
Prevent overlapping matches for the Regex matcher

jonathonherbert:
## What does this change?

This PR removes overlapping matches from the Regex matcher results. Rules are registered in order. Matches produced by later rules supersede matches produced by earlier results.

It's likely we'll need to think carefully about how this might work across matchers, but for now this amends problematic behaviour for the current rule set.

Before:

Screenshot 2020-08-12 at 07 41 17

After:

Screenshot 2020-08-12 at 07 40 19

## How to test

As an example, deploy to CODE, and run Typerighter in CODE Composer against the phrase, 'for many people today, especially for women, packaging and broadcasting your image is a readily monetizable skill'. 'Women' should contain a single match, not two matches, as was previously the case.

We never see a regex rule overlapping another regex rule.


test/scala/matchers/RegexMatcherTest.scala
```diff
@@ -1,29 +1,38 @@
 package matchers
 
 import model._
-import org.scalatest._
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+import com.softwaremill.diffx.scalatest.DiffMatcher._
```
jennygrahamjones, line 6: I've not used this library before. Nice!

jonathonherbert: The diffs without it were ... the worst 😭

jonathonherbert:
@SHession ah, this is a flakey test. We rationalised leaving it a while ago as it's solved by #20, but as this is unlikely to be merged in the very near future we may want solve this separately. So unrelated to this PR, but thanks for raising 👍

PR #62, opened 2020-08-12
[Bugfix]: Apply multiple categories per block, and ensure they're all checked correctly

jonathonherbert:
## What does this change?

#61 exposed a longstanding bug – if we attempt to search across multiple categories in the same block of text, only the last is returned.

This bug was exposed by attempting to use the `blockLevelCheckStrategy` CheckStrategy. Previous to this PR, it only returned the last registered category, throwing away the others.

It should now return all categories correctly.

## How to test

Run locally, or deploy to CODE, and run a check. You should see all categories returning as normal.

Previous to this PR, if we'd used the `blockLevelCheckStrategy` to check results, we would have thrown away all but the last category.

### DIFF
--- a/app/services/MatcherPool.scala
+++ b/app/services/MatcherPool.scala
@@ -12,6 +12,7 @@ import utils.Matcher
 import akka.stream.QueueOfferResult.{Dropped, QueueClosed, Failure => QueueFailure}
 import akka.stream._
 import akka.stream.scaladsl.{Sink, Source}
+import scala.util.Failure
 
 case class MatcherRequest(blocks: List[TextBlock], categoryId: String)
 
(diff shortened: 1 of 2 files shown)


jonathonherbert:
@SHession great idea, and I'll add some more detail to this PR to better explain the change. Any other questions please let me know, a lot of this was written quite some time ago and a refresher is always useful.

app/services/MatcherPool.scala
```diff
@@ -167,23 +167,26 @@ class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, va
   }
 
   private def runValidationJob(job: MatcherJob): Future[(MatcherJob, List[RuleMatch])] = {
-    val jobResults = job.categoryIds.map { categoryId =>
-      matchers.get(categoryId) match {
-        case Some((_, matcher)) =>
-          val eventuallyMatches = matcher.check(MatcherRequest(job.blocks, categoryId))
-          job.promise.completeWith(eventuallyMatches)
```
jonathonherbert, line 174: This is the problem – we're mapping across a list of many category ids, but a job can only be completed once.

So the job is completed with the first category, and returns a result; meanwhile other categories are checked, and their results are discarded.

SHession: Thanks for the explanation, that does make sense 👍

PR #63, opened 2020-08-13
Removes overlapping rules by category

SHession:
## What does this change?

This PR builds on the work completed in #61 to remove overlapping rules from results. It goes on to remove overlapping results at a category and matcher level, rather than just a matcher level. This will prevent rules from overlapping between categories but stops short of preventing overlapping rules at a request level.

The order of the priority for the rules is arbitrarily set to be the category id, this ensures the same match is consistently returned. However, ideally a priority would be specified against a category / state (correct, ambiguous, unambiguous). 

## How to test

Enter a word which has matches against multiple categories such as "Midwest", observe one result is returned consistently.

Users no longer experience issues with rules being highlighted with overlapping matches.

## Images

This bug was initially observed on the Typerighter client demo, highlighting the word midwest under two different categories:
TyperighterBugII

### DIFF
--- a/app/services/MatcherPool.scala
+++ b/app/services/MatcherPool.scala
@@ -8,7 +8,7 @@ import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import play.api.Logging
 import model.{BaseRule, Category, Check, MatcherResponse, RuleMatch, TextBlock}
-import utils.Matcher
+import utils.{Matcher, RuleMatchHelpers}
 import akka.stream.QueueOfferResult.{Dropped, QueueClosed, Failure => QueueFailure}
 import akka.stream._
 import akka.stream.scaladsl.{Sink, Source}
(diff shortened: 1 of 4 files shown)


app/services/MatcherPool.scala
```diff
@@ -171,20 +171,29 @@ class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, va
...
     }
 
-    Future.sequence(eventuallyJobResults).map { results =>
-      val flattenedResults = results.flatten
-      job.promise.completeWith(Future.successful(flattenedResults))
-      (job, flattenedResults)
+    Future.sequence(eventuallyJobResults).map { matchesByCategory =>
+      val sortedMatches = matchesByCategory.sortBy {
+        case (category, _) => category.id
+      }.foldLeft(List.empty[RuleMatch])(
```
SHession, line 187: We sort by category Id to ensure we have a consistent priority of matches.

PR #64, opened 2020-08-17
Add 'ignore' option to the rules sheet

jonathonherbert:
## What does this change?

It'd be nice to be able to deactivate rules on a case by case basis without having to remove them from the active sheet.

This PR reads the ninth row of the rules sheet as 'ignore', and doesn't load the rule if that cell reads 'TRUE'.

## How to test

Run this branch in CODE. None of the rules in the CODE sheet with TRUE ignore cells should appear in the rules endpoint `/rules`.

It's easy to disable troublesome rules at short notice.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -43,8 +43,7 @@ class AppComponents(context: Context, identity: AppIdentity, creds: AWSCredentia
 
   val credentials = configuration.get[String]("typerighter.google.credentials")
   val spreadsheetId = configuration.get[String]("typerighter.sheetId")
-  val range = configuration.get[String]("typerighter.sheetRange")
-  val ruleResource = new SheetsRuleResource(credentials, spreadsheetId, range)
+  val ruleResource = new SheetsRuleResource(credentials, spreadsheetId)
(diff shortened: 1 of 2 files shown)


app/AppComponents.scala
```diff
@@ -43,8 +43,7 @@ class AppComponents(context: Context, identity: AppIdentity, creds: AWSCredentia
 
   val credentials = configuration.get[String]("typerighter.google.credentials")
   val spreadsheetId = configuration.get[String]("typerighter.sheetId")
-  val range = configuration.get[String]("typerighter.sheetRange")
-  val ruleResource = new SheetsRuleResource(credentials, spreadsheetId, range)
+  val ruleResource = new SheetsRuleResource(credentials, spreadsheetId)
```
SHession, line 46: Why is the range no longer coming for the configuration?

jonathonherbert: Ah, yes, should have flagged – because it's tied to the ranges we expect from our columns, which are determined in code. In retrospect, I don't think this should have been parameterised to begin with.

PR #67, opened 2020-08-20
Change replacement type to TextSuggestion

jonathonherbert:
## What does this change?

At the moment, we output replacements as strings. This isn't consistent with our Suggestions model, which represents rule suggestions as a sealed trait of Suggestions. This is designed to accommodate diverse suggestion types over time (for e.g. #12).

This PR replaces the string type with a TextSuggestion type.

## How to test

The API output for matches should now match the TextSuggestion model when replacements are available. You could test with e.g. a local instance of [prosemirror-typerighter](https://github.com/guardian/prosemirror-typerighter).

The automated tests which produce replacements should pass.

### DIFF
--- a/app/model/RegexRule.scala
+++ b/app/model/RegexRule.scala
@@ -21,6 +21,7 @@ case class RegexRule(
 
   def toMatch(start: Int, end: Int, block: TextBlock): RuleMatch = {
     val matchedText = block.text.substring(start, end)
+    val replacementText = replacement.map(replacement => regex.replaceAllIn(matchedText, replacement.text))
     RuleMatch(
       rule = this,
       fromPos = start + block.from,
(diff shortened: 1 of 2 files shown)


app/model/RegexRule.scala
```diff
@@ -21,6 +21,7 @@ case class RegexRule(
 
   def toMatch(start: Int, end: Int, block: TextBlock): RuleMatch = {
     val matchedText = block.text.substring(start, end)
+    val replacementText = replacement.map(replacement => regex.replaceAllIn(matchedText, replacement.text))
```
aug24, line 24: This is a bit unweildy.  If I read it correctly, it goes from Option[TS] to Option[String?] only to be turned back into Option[TS] on line 33.

How about a `replaceAllIn` method on the TextSuggestion object which wraps this functionality returning a new TS?

jonathonherbert: Hm, I'd prefer our `TextSuggestion` to avoid having opinions about regular expressions, but you're right, this is a silly dance – we should at least do it in one step, I think. How about –

```scala
val derivedReplacement = replacement.map(replacement =>
      replacement.copy(text = regex.replaceAllIn(matchedText, replacement.text))
    )
```

aug24: This is definitely better than it was.

If RegexRule objects can have opinions about TextSuggestions, I see no reason the converse shouldn't be true :)

PR #68, opened 2020-09-01
Recover from stream failure

jonathonherbert:
## What does this change?

At the moment, if the stream that manages throughput for the `matcherPool` receives an error, it completes with a failure. This means the whole pool stops working.

This can happen when checks are run whilst the tool is refreshing its matches. Because the categories have been cleared out and not yet replaced with new rules, no categories will exist, and the stream will fail. This is causing boxes to become unresponsive to checks in prod.

This stream behaviour is the default behaviour of akka streams, and `mapAsync` in particular via a failed future. We can use a `SupervisionStrategy` to [prevent this from happening](https://doc.akka.io/docs/akka/current/stream/stream-error.html#errors-from-mapasync).

Another approach would be to emit the error as a value wrapped in a successful future, but we would probably want this behaviour in any case, to ensure that other errors causing the future to fail don't propagate up to the stream.

## How to test

One way to test this is to ask for matches for a category that doesn't exist. This fails a future without the stream, and will cause the `MatcherPool` to stop running jobs.

As above, an easy way produce that result is to repeatedly run checks whilst the tool is refreshing its matches. Hit the refresh button in the `/rules` UI, and immediately run a check. 

The Typerighter service no longer becomes unresponsive when rules are refreshed by users.

## Developer notes

I've refactored the current tests to avoid using ScalaFutures, and removed a test that was intermittently failing as a part of this PR. Happy to bundle in a separate PR if needed.

### DIFF
--- a/app/services/MatcherPool.scala
+++ b/app/services/MatcherPool.scala
@@ -62,8 +62,15 @@ class MatcherPool(val maxCurrentJobs: Int = 8, val maxQueuedJobs: Int = 1000, va
 
   private val matchers = new ConcurrentHashMap[String, (Category, Matcher)]().asScala
   private val eventBus = new MatcherPoolEventBus()
+
+  // This supervision strategy resumes the stream when `mapAsyncUnordered`
+  // emits failed futures. Without it, the stream fails when errors occur. See
+  // https://doc.akka.io/docs/akka/current/stream/stream-error.html#errors-from-mapasync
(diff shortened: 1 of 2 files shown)


SHession (commented):
Looks like you missed a ScalaFutures on :268

jonathonherbert:
Thanks for spotting @SHession, just corrected 👍

SHession (approved):
I've reviewed the code changes and rationale and they make sense. I've also tested this locally and on CODE and found that locally was able to recover from the rule refreshing issues but CODE was not (as expected).

PR #69, opened 2020-09-01
Remove first row from sheet on ingest

jonathonherbert:
## What does this change?

This PR removes the first row from the google sheet on ingest – it's the heading, and right now we remove it by adding an intentional error to the 'regex' cell 😅 

## How to test

Spin this up in e.g. CODE. You should no longer see the table header of the CODE rules sheet being ingested.


SHession (approved):
Reviewed and tested locally, LGTM

PR #70, opened 2020-09-17
Fix mapping of getCurrentCategories to return category ids not matcher ID

tjsilver:
## What does this change?
This work is related to [this PR](https://github.com/guardian/prosemirror-typerighter/pull/136) and fixes a bug that meant that the matcher IDs were returned instead of the category IDs.

## How to test
Run branch locally with the same branch from `prosemirror-typerighter` and see the category IDs in the response.

### DIFF
--- a/app/controllers/ApiController.scala
+++ b/app/controllers/ApiController.scala
@@ -26,7 +26,7 @@ class ApiController(
             val response = MatcherResponse(
               matches = matches,
               blocks = check.blocks,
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
👍

tjsilver:
[Check all categories by default](https://trello.com/c/XJmMnbn9/402-check-all-categories-by-default)

PR #71, opened 2020-09-17
Fails rule ingestion if any rules fail

jonathonherbert:
## What does this change?

Fails rule ingestion if any rules fail, to encourage us to provide well-formed rules. We report via the UI if this happens via the 'refresh' bottom.

## How to test

Induce an error in a rule – for example by adding a single { in the 'pattern' field. Visit the `/rules` endpoint. When pressing 'refresh' , you should see that
- no rules are added
- the error is clearly enumerated

Please contact me directly if you'd like links to the code environment or the rules sheet.
 
We only ingest new rulesets if all of the incoming rules are well-formed.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -86,12 +86,17 @@ class AppComponents(context: Context, identity: AppIdentity, creds: AWSCredentia
     */
   def initialiseMatchers: Future[Unit] = {
     for {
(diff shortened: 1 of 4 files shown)


SHession (approved):
Code changes look good, I've only tested the happy path (no errors) as I don't have permission to access to the CODE sheet.

jennygrahamjones (approved):

SHession:
Since tested, works correctly,

PR #72, opened 2020-09-17
Add ID to rules and be specific in error message when ID is not present

jonathonherbert:
## What does this change?

Adds a ruleId to ingested rules via a column in the sheet. We add UUIDs to this column.

We issue an error if a rule doesn't have an ID.

## How to test

Visit the `/rules` endpoint. You should see that rules are now added with IDs. 

When pressing 'refresh' , you should see that removing an ID from the sheet should give an error in the Typerighter UI.

Please contact me directly if you'd like links to the code environment or the rules sheet.

We have a unique id per rule, that is stably tied to that rule.


SHession (approved):
LGTM

PR #73, opened 2020-09-18
Handle thrown errors correctly in matchers

jonathonherbert:
## What does this change?

Imagine my horror upon discovering that exceptions thrown in the `check` method in matchers were never caught, and instead simply left the job promise open!

This PR adds an execution context to the check function, provided by the MatcherPool, to ensure check work is executed in its proper context.

Apart from the implications for concurrency, this is necessary because if error are thrown in `check`, rather than passed as values (`Future.failed` etc.) and remain uncaught, the calling thread stops executing code, and never completes the promise that has been passed back to the stream.

By wrapping `check` in a Future block that uses the provided execution context, we can ensure that the error propagates as a failed future rather than halting execution.

## How to test

The automated test should give you an idea of how to trigger this – force a matcher to throw an error. Adding the following rule to e.g. the CODE rules sheet and checking text that contains the word 'plant' 🌱  should do it:

<details><summary>Bad rule! Bad!</summary>

regex | \b(?i)plan-? ?([A-Z])\b | plan $2 | 3ff200 | Style guide and names |   | General: plan B, etc | General | FALSE |   | d7541ff7-063a-41de-80d5-74fd8d7c3e8c | 
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

</details>

Before, your check would never complete. Now it fails fast. Inspecting the network requests, you should see an error for the offending block.

We no longer see requests time out when errors are thrown by our matchers.


SHession (approved):
Tested and reviewed locally LGTM!

PR #74, opened 2020-09-18
Add timeout to matcher checks

jonathonherbert:
## What does this change?

Adds a timeout to each call to `check`, configured by the `MatcherPool` This should ensure that users don't have to wait for their socket to time out (30 seconds? A minute?) before they see checks fail.

The initial, misiguided response to a problem that was solved by #73 – although the problems we saw with odd regexes didn't end up being caused by a timeout, I think it's worth adding one as long as we've a test to ensure it works, as it's likely we'll eventually run into a problem with long-running and possibly pathological calls to `check` for some matcher.

## How to test

We don't have a way of inducing a timeout on a single check with runtime configuration for now. You could manually alter the code in this branch to never complete a future, and observe the results.

The automated test, which effectively does the same thing, should pass.

If we introduce rules that cause checks to time out, users (and us!) are alerted early.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -20,6 +20,7 @@ import router.Routes
 import rules.SheetsRuleResource
 import services._
 import utils.Loggable
+import play.api.libs.concurrent.DefaultFutures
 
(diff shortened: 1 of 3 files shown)


SHession (approved):
Reviewed and tested locally LGTM!

PR #76, opened 2020-09-21
Add LanguageTool to our pool of matchers, and enable its default rules via the sheet

jonathonherbert:
## What does this change?

Add LanguageTool to our pool of matchers, and enable its default rules via our sheet.

Screenshot 2020-09-21 at 16 03 14

## How to test

Working locally, or in CODE, take a look in CODE at the palette of LanguageTool rules in the CODE sheet. Enable or disable rules and refresh the rules via UI available at the `/rules`endpoint. Do they behave as expected? ([LanguageTool's live demo](https://languagetool.org/) is a good point of comparison.)

Anyone with access to the relevant sheet can enable or disable LanguageTool rules.

We'll need to revise almost immediately to play nicely with #75. 

## Notes

The rules produced by LanguageTool don't correspond to our usual way of handling categories, which we strongly suspect is due for revision.

With RegexRules, there's a 1-1 mapping between category and matcher. For LT that relationship is broken – there's a 1-1 mapping between a category and matcher, but a many-1 mapping between a _rule_ and a category – so different rules within the same matcher 'category' can produce different categories when matches are made.

This is useful for now, as otherwise we'd see 'Default LanguageTool rules' in our category output for LT matches, which is no use to anybody at present – but it's not obvious to anyone familiar with the way rules are structured.

In future, the Category entity is likely to be replaced with a Tag entity with a many-many relationship, which should solve these sorts of problems.


paperboyo:
💎!

SHession (commented):
This looks good but I will wait until these changes have been rebased against the rule ingestion work before testing.

jonathonherbert:
@SHession this PR has grown a bit to incorporate a refactor that was necessary following #75 – happy to pair on the changes tomorrow if needed.

app/rules/BucketRuleManager.scala
```diff
@@ -0,0 +1,61 @@
...
+          rulesMeta.getLastModified
+        }
+    }
+
+    def logOnError[T](op: => T): Either[Exception, T] = {
+        try {
+          Right(op)
+        } catch {
+          case e: Exception => {
+                log.warn(e.getMessage)
```
SHession, line 56:
```suggestion
                log.error(e.getMessage)
```

SHession: This error log could also give some more helpful information

SHession (commented):
I've tested this locally and it appears to work well although I can't see the Language Tool rules listed on the Rules page.

app/rules/BucketRuleManager.scala
jonathonherbert: Good point, just added.

jonathonherbert:
@SHession I can spot the LT rules, but they're right at the bottom! If you confirm, we can definitely do more to improve the UI here, but that might be for another PR – for example, a filterable list of rules, and we don't include the matcher name in this list, either. E.g 
Screenshot 2020-09-23 at 14 26 47

SHession:
> @SHession I can spot the LT rules, but they're right at the bottom! If you confirm, we can definitely do more to improve the UI here, but that might be for another PR.

I have since spotted them, I hadn't been able to find them due to searching for the category of `languagetool`. As explained by @jonathonherbert they are not categories in this way. As the way we categorise is likely to change in the future it's not worth worrying about for now.

SHession (approved):
Tested the latest changes and reviewed the code. All looks good to me!

jonathonherbert:
Amazing! Now I've just got to figure out why boxes aren't coming up ...

PR #77, opened 2020-09-23
Add matcher type to rules

tjsilver:
Co-authored-by: Jonathon Herbert <jonathon.herbert@guardian.co.uk>

## What does this change?
At the moment, our rules don't include the name of their matcher. This will be useful in determining whether rules from different matchers have different telemetry profiles.

This PR adds the matcher type to the match that's produced to pass it to the client as a part of the response. We then pass the matcher type to the telemetry data in `prosemirror-typerighter` [in this PR](https://github.com/guardian/prosemirror-typerighter/pull/139).

## How to test
Run the branch locally along with the client (including cookie from Composer). Run 'Check document' and see the `matcherType` in the responses.

We're able to determine which matcher produced a match in the response.

### DIFF
--- a/app/matchers/LanguageToolMatcher.scala
+++ b/app/matchers/LanguageToolMatcher.scala
@@ -8,7 +8,7 @@ import org.languagetool.rules.spelling.morfologik.suggestions_ordering.Suggestio
 import org.languagetool.rules.{CategoryId, Rule => LanguageToolRule}
 import play.api.Logging
 import services.MatcherRequest
-import utils.Matcher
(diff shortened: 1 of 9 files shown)


tjsilver:
[Add matcher to telemetry tags](https://trello.com/c/6zoiMU8a/407-add-matcher-to-telemetry-tags)

jonathonherbert:
Tested locally and works as expected 💯 – please note, I've rebased this from master, so if you'd like to add anything don't forget to force pull first!

jonathonherbert (approved):

PR #78, opened 2020-09-24
Refactor to allow multiple matchers per category

jonathonherbert:
## What does this change?

Previous to this PR, categories had a 1-1 mapping with matchers. This is problematic if we'd like to share a category across matchers, which we'll want to do shortly as LanguageTool is used more widely in the project.

For example, we might like to have both a RegexMatcher and a LanguageTool matcher that output the 'Style Guide' category.

This PR refactors matcherPool and the provisioning services to allow multiple matchers per category.

Generated matches are still sorted and ordered by categories on a per-job basis to provide deterministic behaviour when handling overlapping rules.

## How to test

This is a no-op. The automated tests should pass as usual.

We can add multiple matchers per category in future PRs.


app/model/BaseRule.scala
```diff
@@ -14,11 +29,119 @@ trait BaseRule {
...
 
 object BaseRule {
-  def toJson(rule: BaseRule): JsObject = Json.obj(
-    "id" -> rule.id,
-    "category" -> rule.category,
-    "description" -> rule.description,
-    "suggestions" -> rule.suggestions,
-    "replacement" -> rule.replacement
-  )
+  implicit val writes: Writes[BaseRule] = Json.writes[BaseRule]
```
jonathonherbert, line 32: We convert our rules to a sealed trait here to allow implicit serialisation entities that inherit the `BaseRule` trait – see [the Scala docs](https://www.playframework.com/documentation/2.8.x/ScalaJsonAutomated#Requirements) for why this is needed.

app/model/RuleMatch.scala
```diff
@@ -30,19 +30,7 @@ object RuleMatch {
...
-    "matchedText" -> ruleMatch.matchedText,
-    "message" -> ruleMatch.message,
-    "replacement" -> ruleMatch.replacement,
-    "shortMessage" -> ruleMatch.shortMessage,
-    "suggestions" -> ruleMatch.suggestions,
-    "markAsCorrect" -> ruleMatch.markAsCorrect,
-    "matchContext" -> ruleMatch.matchContext,
-    "matcherType" -> ruleMatch.matcherType
-  ))
+  implicit val writes: Writes[RuleMatch] = Json.writes[RuleMatch]
```
jonathonherbert, line 33: We're now able to serialise a `BaseRule` via its implicits, so this manual conversion is no longer needed.

app/services/MatcherPool.scala
```diff
@@ -69,8 +69,8 @@ class MatcherPool(
   val checkTimeoutDuration: FiniteDuration = 5 seconds
 )(implicit ec: ExecutionContext, implicit val mat: Materializer) extends Logging {
   type JobProgressMap = Map[String, Int]
-
-  private val matchers = new ConcurrentHashMap[String, (Category, Matcher)]().asScala
+  type MatcherId = String
+  private val matchers = new ConcurrentHashMap[MatcherId, Matcher]().asScala
```
jonathonherbert, line 73: NB: instead of mapping from a category to a matcher, we give matchers a unique ID and use that as the key. The filtering code in `runValidationJob` below determines which matches are eligible for which jobs.

SHession (approved):
Reviewed and tested this code locally and refactored some functions with @jonathonherbert to conform to the idea of "parse don't validate".

app/rules/SheetsRuleManager.scala
```diff
@@ -47,8 +47,8 @@ class SheetsRuleManager(credentialsJson: String, spreadsheetId: String, matcherP
     }
   }
 
-  private def getRegexRules()(implicit ec: ExecutionContext): Either[List[String], List[RegexRule]] = {
-    getRulesFromSheet("regexRules", "A:N", getRegexRuleFromRow)
+  private def getPatternRules()(implicit ec: ExecutionContext): Either[List[String], List[BaseRule]] = {
+    getRulesFromSheet("regexRules", "A:N", getRuleFromRow)
```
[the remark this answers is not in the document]
jonathonherbert: Great spot 👍  I've added some context in function comments re: PatternRules to make what's going on a bit clearer, we're anticipating a future case here.

app/rules/SheetsRuleManager.scala
```diff
@@ -86,41 +86,51 @@ class SheetsRuleManager(credentialsJson: String, spreadsheetId: String, matcherP
...
+        case (_, Some("TRUE"), _) => Success(None)
+        case (None, _, _) => Failure(new Exception(s"no id for rule"))
+        case (Some(id), _, _) if id.length == 0 => Failure(new Exception(s"empty id for rule"))
+        case (Some(id), _, "regex") => Success(Some(getRegexRule(
+            id,
+            row(1).asInstanceOf[String],
+            row(2).asInstanceOf[String],
+            row(3).asInstanceOf[String],
+            row(4).asInstanceOf[String],
+            row.lift(6).asInstanceOf[Option[String]]
```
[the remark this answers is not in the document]
jonathonherbert: Yeah, it felt gross writing those out, your idea makes things much clearer. I've added an object to keep them in – hopefully that makes sense!

jennygrahamjones (approved):

PR #80, opened 2020-10-06
Add LanguageTool rules from our rules sheet

jonathonherbert:
**NB:** please review and merge #81 first, which'll slim down the diff a bit!

## What does this change?

[Trello card](https://trello.com/c/jG3f0kCp/459-add-custom-languagetool-rules-to-our-google-sheet)

Adds the ability to add LanguageTool rules from our rules sheet. For an example of the way these rules are constructed within LanguageTool directly, see e.g. [this xml](https://raw.githubusercontent.com/languagetool-org/languagetool/master/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml).

These rules look much the same as the regex rules. The key differences:
- the 'pattern' section accepts an XML description of everything that's possible to add to a LanguageTool rule definition, within the `<rule / >` tag.
- the 'replacement' field isn't used – the XML takes care of that.

## How to test

Add a new rule to the CODE sheet and test this, either locally or in CODE.

<details><summary>An example rule, taken from the LT defaults.</summary>

Rule type | Pattern | Replacement? | Name | Category |   | Description | Tags (separate with commas) | Ignore? | Notes | ID
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
lt | ```<pattern> <token postag="CD"/> <token postag="NNS"><exception regexp="yes">centuries\|decades\|years\|months\|days\|hours\|minutes\|seconds\|stars</exception></token> <token>or</token> <marker> <token>less</token> </marker> </pattern> <message>Did you mean <suggestion>fewer</suggestion>? The noun \2 is countable.</message> <short>Grammatical error</short> <example correction="fewer">Ten items or <marker>less</marker></example> <example>It typically takes 30 seconds or <marker>less</marker></example> <example>I would only give that hotel 3 stars or <marker>less</marker>.</example>``` |   | Fewer / less | Grammar |   | An example LT rule to get us started. | Grammar | FALSE |   | FEWER_LESS_CUSTOM 

</details>

This rule should match the `less` part of the phrase, `Three mistakes or less`.

Bear in mind that adding this rule in the CODE sheet will break other users of the CODE sheet, which includes DEV! I found this out to my cost whilst testing 😬 – it's not catastrophic, but worth letting other devs know what you're up to.

We can add LanguageTool custom rules via that spreadsheet. 

They report correctly in the UI, including descriptions, ranges, and suggestions.

## Notes

WRT

> the 'pattern' section accepts an XML description of everything that's possible to add to a LanguageTool rule definition, within the `<rule / >` tag.

this may prove to be too restrictive over time, as it means user defined rules don't get access to the rule attributes, or any tags that might be adjacent or contain them, e.g.`rulegroup`. The advantage of this approach is that we can keep the ID, description properties etc. in the sheet, which makes the rules easier to audit at a glance, but we can change tack if it becomes clear our rule authors require more flexibility.

### DIFF
--- a/test/resources/logback.xml
+++ b/test/resources/logback.xml
@@ -17,13 +17,13 @@
 	<!-- uncomment and set to TRACE to log all HTTP requests -->
 	<!--<logger name="io.gatling.http.engine.response" level="TRACE" />-->
 
-	<!--<root level="WARN">-->
-		<!--<appender-ref ref="CONSOLE" />-->
-	<!--</root>-->
-	<!--<root level="TRACE">-->
-		<!--<appender-ref ref="FILE" />-->
-	<!--</root>-->
-	<!--<root level="DEBUG">-->
-		<!--<appender-ref ref="FILE" />-->
-	<!--</root>-->
+	<root level="WARN">-->
+		<appender-ref ref="CONSOLE" />
+	</root>
+	<root level="TRACE">
+		<appender-ref ref="FILE" />
+	</root>
+	<root level="DEBUG">
+		<appender-ref ref="FILE" />
+	</root>
 </configuration>
--- a/app/matchers/LanguageToolMatcher.scala
+++ b/app/matchers/LanguageToolMatcher.scala
@@ -52,9 +52,7 @@ class LanguageToolFactory(
     // Add custom rules
     val ruleErrors = applyXMLRules(instance, ruleXMLs)
 
-
     instance.enableRuleCategory(new CategoryId(category.id))
-    println(instance.getAllActiveRules)
 
     (new LanguageToolMatcher(category, instance), ruleErrors)
   }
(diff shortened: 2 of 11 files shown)


test/resources/logback.xml
```diff
@@ -17,13 +17,13 @@
...
-	<!--<root level="WARN">-->
-		<!--<appender-ref ref="CONSOLE" />-->
-	<!--</root>-->
-	<!--<root level="TRACE">-->
-		<!--<appender-ref ref="FILE" />-->
-	<!--</root>-->
-	<!--<root level="DEBUG">-->
-		<!--<appender-ref ref="FILE" />-->
-	<!--</root>-->
+	<root level="WARN">-->
```
jonathonherbert, line 20: This allows our tests to log, which should probably be a default.

sihil (approved):
This mostly looks great. I came across a couple of minor things whilst testing.
 1. The rule type isn't displayed at the current rules endpoint, not a big deal but could be helpful and a trivial change
 1. When I make the current rules full screen I get a large margin on the left as the container is too narrow for the content, would also be lovely to fix this but not really related to this PR.

jonathonherbert:
@sihil worthy changes both, I'll add 1. in this PR. 2. is probably something to leave until we move this to a management service outright, as there's loads we'll want to do as a result of that migration and I suspect this view won't survive.

PR #81, opened 2020-10-06
Remove colour from the Category model – we don't need it any more

jonathonherbert:
## What does this change?

Since we [updated our traffic light model](https://github.com/guardian/prosemirror-typerighter/pull/95), we haven't used our Category colours. Let's remove them. 

## How to test

This is a no-op for our [only known consumer](https://github.com/guardian/prosemirror-typerighter), but in an API response that includes a match, we should no longer see a colour attribute as part of the `category` part of the `rule` definition.

Fewer LOC? 😄


sihil (approved):
LGTM :shipit:

PR #82, opened 2020-10-07
Make 'before' and 'after' context available in match

jonathonherbert:
## What does this change?

Adds a 'before' and 'after' property to the `Match` interface, to allow our interface to programmatically reconstruct the context of a match. 

Before this, we had to manually parse out the matched text (example string: `What a [misteak] to make`), which was unreliable.

## How to test

Run a check. `before` and `after` properties should be present in the match data when matches are made.

The automated tests cover this output, and should pass.
 
We're able to programmatically reconstruct the context of a match in tooling consuming our API, without parsing strings :)))


SHession (approved):

test/scala/matchers/RegexMatcherTest.scala
```diff
@@ -67,12 +69,13 @@ class RegexMatcherTest extends AsyncFlatSpec with Matchers {
...
                        |""".stripMargin.replace("\n", "")
 
-    val matchText = """
+    val before = """
                        |123456789
                        | 123456789 123456789 123456789
                        | 123456789 123456789 123456789
                        | 123456789 123456789 123456789
-                       | [text]
+                       | """.stripMargin.replace("\n", "")
```
aug24, line 77:
```suggestion
                       | """.stripMargin.mkString("")
```

app/model/BaseRule.scala
```diff
@@ -51,18 +51,20 @@ case class RegexRule(
 
   def toMatch(start: Int, end: Int, block: TextBlock): RuleMatch = {
     val matchedText = block.text.substring(start, end)
-
+    val (before, after) = Text.getSurroundingText(block.text, start, end)
```
aug24, line 54: If we can find a wording that is clearer, I'd prefer it?  Leading / trailing?

aug24 (changes requested):

app/model/BaseRule.scala
jonathonherbert: Changed to precedingText/subsequentText, thanks for pointing out and to @twrichards for the suggestion 👍

PR #83, opened 2020-10-07
Derive categories from rules, not matchers

jonathonherbert:
## What does this change?

In `MatcherPool`, the entity which provides an interface to run a `check` against multiple `matcher` instances, we've a 1-1 mapping between a category and a matcher.

This is restrictive, but worse – it's incorrect! We assign a 'category' to our languageTool matchers, but these matchers have rules that have their own opinions about categories. So a check against an lt instance supposedly responsible for category A may give a response containing category B.

As a result, there's an inconsistency where matchers will report being responsible for certain categories, and deliver rules that report other categories. This causes confusion in tooling downstream.

This PR refactors the `getCategory` interface in `Matcher` to report several categories, and in both `regexMatcher` and `languageTool` matchers changes the implementation to derive the category list from the rules they contain.

There's a broader question as to how we map from taxonomic entities from matchers that is still unanswered, but this PR at least ensures we're honest in our reporting!

## How to test

The automated tests should pass – I've added one to ensure LanguageTool is reporting correctly.

We can test manually this by verifying that a bug in [prosemirror-typerighter] is fixed. Before, the fact that categories were misreported meant that LanguageTool rules would accumulate in the client, because the client clears matches by category, and the categories reported by the matcherResponse root and the matches themselves never added up:

categories-ko

This should no longer happen:

categories-ok

The `MatcherResponse` always reports the categories it has checked correctly.


app/controllers/ApiController.scala
```diff
@@ -26,7 +26,7 @@ class ApiController(
             val response = MatcherResponse(
               matches = matches,
               blocks = check.blocks,
-              categoryIds = check.categoryIds.getOrElse(matcherPool.getCurrentCategories.map { _._2.id })
+              categoryIds = matches.map(_.rule.category.id).toSet.toList
```
sihil, line 29: I think `distinct` is more elegant here:
```suggestion
              categoryIds = matches.map(_.rule.category.id).distinct
```
I think the order of first instance is maintained, but since a set won't maintain any order I doubt you care.

app/matchers/LanguageToolMatcher.scala
```diff
@@ -75,8 +73,11 @@ class LanguageToolFactory(
       case Success(_) => None
       case Failure(e) => {
         logger.error(e.getMessage(), e)
-        Some(e.getMessage())
+        Some(e)
       }
+    } match {
+      case Nil => Right(())
+      case errors => Left(errors)
```
sihil, line 80: This isn't very elegant but can't immediately think of ways to improve it. I suspect using `Validation` might help - I wonder if ZIO has such a thing or if we'd need to lean on cats.

app/services/MatcherPool.scala
```diff
@@ -91,7 +91,7 @@ class MatcherPool(
     */
   def check(query: Check): Future[List[RuleMatch]] = {
     val categoryIds = query.categoryIds match {
-      case None => getCurrentCategories.map { case (_, category, _) => category.id }
+      case None => getCurrentCategories.flatMap { case (_, categories, _) => categories.map(_.id) }
```
sihil, line 94: Can you end up with duplicate category ids here if multiple matchers check for the same category? More interestingly why does getCurrentCategories return something other than categories?

sihil: Possibly the `CategoryIds` type could be a Set instead of a List.

app/services/MatcherPool.scala
```diff
@@ -215,6 +215,9 @@ class MatcherPool(
     }
   }
 
+  private def doesMatcherServeAllCategories(categoryIds: List[String], matcher: Matcher) =
```
sihil, line 218: This actually seems to be asking if the matcher serves any category in the provided list, rather than the all in the function name? `canMatcherGenerateResultsForAnyCategory`?

app/services/MatcherPool.scala
```diff
@@ -136,9 +136,9 @@ class MatcherPool(
     matchers.clear
   }
 
-  def getCurrentCategories: List[(String, Category, Int)] = {
+  def getCurrentCategories: List[(String, Set[Category], Int)] = {
```
sihil, line 139: This is begging to be renamed and return a `List[NewBettererType]`

app/services/RuleProvisionerService.scala
```diff
@@ -32,21 +32,18 @@ class RuleProvisionerService(
       case (category, rules) => {
```
sihil, line 32: `category` is no longer used here. Do you still need to do the `groupBy` on the line above or can it be simplified?

app/services/RuleProvisionerService.scala
```diff
@@ -79,4 +76,14 @@ class RuleProvisionerService(
   def scheduleUpdateRules(scheduler: Scheduler): Unit = {
     scheduler.scheduleWithFixedDelay(0.seconds, 1.minute)(this)
   }
+
+  private def addLTMatcherToPool(matcherPool: MatcherPool, xmlRules: List[LTRuleXML], defaultRules: List[String] = Nil) = {
+    languageToolFactory.createInstance(xmlRules, defaultRules) match {
+      case Right(matcher) => matcherPool.addMatcher(matcher)
+      case Left(errors) => {
+        logger.error(s"Could not create languageTool instance from ruleResource: ${errors.size} errors found")
+        errors.foreach(e => logger.error(e.getMessage(), e))
```
sihil, line 85: This is going to log the exception message twice, perhaps it could be a static string of "rule resource error" to make it easy to marry up with the line above.

sihil (approved):
Functionally fine with the exception of possible duplicate category Ids in lists. A handful of other minor suggestions.

app/matchers/LanguageToolMatcher.scala
jonathonherbert: Yeah, it's a bit frustrating. Interested to see if a common pattern for effects with failure modes helps resolve

app/services/MatcherPool.scala
jonathonherbert: The signature of getCurrentCategories is pretty 'orrid – really it's built to support the rules view. This probably isn't the place to refactor, as it'll mean a rejig of the view, but it sure could use one.

A set rather than a list is sensible.

app/services/MatcherPool.scala
jonathonherbert: Agreed. Hm, maybe we could just refactor in this PR. I'll give it a go.

app/services/MatcherPool.scala
jonathonherbert: The `allCategories` in this name refers to the requested categories, not the matcher categories – that is to say, can the given matcher serve all of the categories in `categoryIds`? There's probably a better name out there – e.g. `doesMatcherServeCategoryIds`?

app/services/RuleProvisionerService.scala
jonathonherbert: We can make category a no-op, but because we produce a single regex matcher per category, we still need the groupBy.

app/services/MatcherPool.scala
jonathonherbert: Done.

app/services/RuleProvisionerService.scala
```diff
@@ -81,8 +81,9 @@ class RuleProvisionerService(
     languageToolFactory.createInstance(xmlRules, defaultRules) match {
       case Right(matcher) => matcherPool.addMatcher(matcher)
       case Left(errors) => {
-        logger.error(s"Could not create languageTool instance from ruleResource: ${errors.size} errors found")
-        errors.foreach(e => logger.error(e.getMessage(), e))
+        val logPrefix = "RuleProvisionerService error: "
+        logger.error(s"${logPrefix} could not create languageTool instance from ruleResource: ${errors.size} errors found")
+        errors.foreach(e => logger.error(s"${logPrefix} ${e.getMessage()}", e))
```
sihil, line 86: You don't need e.getMessage here. It'll simply duplicate the data from the exception.

app/services/MatcherPool.scala
```diff
@@ -136,12 +136,11 @@ class MatcherPool(
...
-    val matchersAndCategories = matchers.values.map { matcher =>
-      (matcher.getType, matcher.getCategories, matcher.getRules.length)
-    }.toList
-    matchersAndCategories
-  }
+  def getCurrentMatchers: List[Matcher] =
+    matchers.values.toList
+
+  def getCurrentCategories: Set[Category] =
+    matchers.values.flatMap { _.getCategories }.toSet
```
sihil, line 143: Much simpler - definitely feels like the right decision.

sihil (approved):

app/services/RuleProvisionerService.scala
jonathonherbert: Ah, sorry, misunderstood. Corrected.

PR #87, opened 2020-10-13
Add performance logging every time we run a matcherPool check

jonathonherbert:
## What does this change?

Adds some logging on the server to give us indexed access to the time it takes to run each request.

We should receive enough information in the context provided in these logs to be able to identify the document, the user concerned, the categories used, and the time it took to run the check.

## How to test

Run a check with this service running in CODE. You should see the relevant metrics appear in the logs.

Screenshot 2020-10-13 at 18 11 32

We're missing documentIds and categoryIds – the former because we don't yet pass that through to Typerighter on the client, and the latter because we aren't asking for specific categories (yet).

We know whether our tool is performing timely checks, and have enough information to narrow down the users affected when it's not performing as expected.

### DIFF
--- a/app/services/MatcherPool.scala
+++ b/app/services/MatcherPool.scala
@@ -16,6 +16,7 @@ import akka.stream._
 import akka.stream.scaladsl.{Sink, Source}
 import play.api.libs.concurrent.Futures
 import play.api.Logging
+import utils.Timer
 
 case class MatcherRequest(blocks: List[TextBlock])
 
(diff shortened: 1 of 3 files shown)


app/services/MatcherPool.scala
```diff
@@ -106,6 +107,8 @@ class MatcherPool(
 
     val eventuallyResponses = jobs.map(offerJobToQueue)
 
+    val totalCharsForCheck = query.blocks.foldLeft(0)((acc, block) => acc + block.text.size)
+
```
SHession, line 111: I assume the changes in the file are for some future PR?

app/utils/Timer.scala
```diff
@@ -0,0 +1,47 @@
...
+    val markers = Markers.appendEntries((
+      Map(
+        "taskName" -> taskName,
+        "durationInMs" -> durationInMs
+      )
+    ).asJava)
+    markers.add(additionalMarkers)
+
+    val message = s"Task $taskName complete in ${durationInMs}ms"
+    logger.info(message)(markers)
```
SHession, line 45: `nit` This log level could perhaps be better set to `debug` to avoid polluting the logs.

SHession (approved):
Reviewed code and test locally, happy the PR works as expected.

app/services/MatcherPool.scala
jonathonherbert: Oops, good spot – yes, they're for our slow log, which will come later.

PR #89, opened 2020-10-14
Adds cloudfront to cfn

SHession:
## What does this change?

This PR adds Cloudfronts to the project cloud formation. This should reduce latency for users in other countries. 

## How to test

This should be a no-op from a user perspective.


sihil (approved):
LGTM, having paired on the impl :shipit:

jonathonherbert:
🎉 Out of interest, what do we need the bucket for?

sihil:
Storing access logs. Copied from composer and thought it was a sane thing to include.

PR #94, opened 2020-10-27
Add rule ingestion alarm

SHession:
## What does this change?

This PR provides alerting for a failure to ingest the latest rules. This is done through the usage of custom CloudWatch metrics for the rules not being found. If no rules are found consistently for three minutes, an alert will trigger, indicating that there is an issue. This is likely to happen when either the rules are missing from the S3 bucket, or there is an S3 outage of some kind. 

## How to test

Deploy the changes to code (ensure the alarm is created, which it is not normally for CODE), delete the rules from the S3 bucket, wait three minutes, observe the alarm has triggered. 

We are notified when there is an issue with the Typerighter rule ingestion, allowing us to resolve potential issues quickly.


app/services/RuleProvisionerService.scala
```diff
@@ -13,11 +13,14 @@ import rules.BucketRuleManager
...
 import model.LTRuleXML
+import utils.CloudWatchClient
+import utils.Metrics
 
 class RuleProvisionerService(
   bucketRuleManager: BucketRuleManager,
   matcherPool: MatcherPool,
-  languageToolFactory: LanguageToolFactory
+  languageToolFactory: LanguageToolFactory,
+  cloudWatchClient: CloudWatchClient
```
SHession, line 23: There may be a nice way of providing access to the cloudwatch client, incase we would like to use it on other places. @sihil

typerighter.cfn.yaml
```diff
@@ -426,3 +446,26 @@ Resources:
           Bucket: !GetAtt CloudfrontBucket.DomainName
           IncludeCookies: 'false'
           Prefix: ''
+
+  RuleProvisionerAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdAlarms
```
SHession, line 452: This condition prevents the alarm from being created on CODE, remove this if you wish to test.

jonathonherbert (approved):
LGTM, I can see metrics in CW and an alarm triggered by this CODE 👍

PR #95, opened 2020-10-30
Allow rules and rule groups in sheet

tjsilver:
## What does this change?
Currently, when users create a new LanguageTool custom rule we ingest the rule and wrap in a `<rule>` tag. This however doesn't allow users to add custom rule groups. To allow this, we now ingest the rule pre-wrapped in either a root `<rule>` or `<rulegroup>` tag, and add the `id` and `name` attributes into the root tag.

NOTE: The existing custom LT rules will need to be wrapped in `<rule>` or `<rulegroup>` at the same time that this PR is merged.

## How to test
Run the branch locally and test. The CODE sheet is currently set up with wrapped rules for testing. 

Test the two rulegroups and the individual rules against text with incorrect dates, ellipses and references to 'woman' or 'women' as adjectives to test the new rulegroups, e.g. "The 5th of October, January 3rd, 5th August 2017… three…four…five. There are women electricians."

Rule creators will be able to drop both rules and rule groups into the rules spreadsheet.

Existing custom rules will need to be wrapped in `<rule>` or `<rulegroup>` tags in order to work in future.

### DIFF
--- a/app/matchers/LanguageToolMatcher.scala
+++ b/app/matchers/LanguageToolMatcher.scala
@@ -16,7 +16,7 @@ import model.Category
 import org.languagetool.rules.patterns.PatternRuleLoader
 import org.languagetool.rules.patterns.PatternRule
 import org.languagetool.rules.patterns.AbstractPatternRule
-import scala.xml.XML
+import scala.xml.{XML, MetaData, Attribute, Null, Text}
 import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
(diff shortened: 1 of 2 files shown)


tjsilver:
[Add custom LanguageTool rules to our google sheet, pt.2 – allow ruleGroups](https://trello.com/c/2dBrmcSf/481-add-custom-languagetool-rules-to-our-google-sheet-pt2-allow-rulegroups)

jonathonherbert (approved):
Great to see tests for the new cases 👍 

Are we looking at a brief period of downtime whilst we make this change? I'd be v. happy to help make this backwards compatible if that'd help, but it might be that it's not worth the effort!

jonathonherbert:
NB: concluded that downtime for just a few rules is low impact, so proceeding for now. Additional tests most welcome in another PR 👍

PR #96, opened 2020-11-02
Improve error handling for LanguageTool xml parsing

jonathonherbert:
**NB**: Depends on #95. Please, gentle reader – review/approve/merge that first!

## What does this change?

This PR improves error handling for LanguageTool xml parsing. Before, LanguageTool rules that contained invalid XML would fail silently to the user, and only emit to logs. Now, the frontend sees the error, correlated with the rule id: 

Screenshot 2020-11-02 at 18 36 42

## How to test

Write a duff rule in the rules sheet (find an example in the unit tests), and hit 'refresh' on the rule management page at `/rules`. You should see a suitable error message.

Rule-makers have a clearer idea about what's wrong when they're editing LanguageTool xml.

In the messaging, we make it clear that when we're parsing rule XML, if a single rule fails, others for that category will also fail. This is because we create matchers on a per-category basis.

It'd be quite a lot more work to be resilient to errors on a per-rule basis, and that behaviour should be redundant once we move to a separate rule management service, so we should avoid the effort if possible.

### DIFF
--- a/app/controllers/RulesController.scala
+++ b/app/controllers/RulesController.scala
@@ -31,7 +31,10 @@ class RulesController(
       maybeRules.left.map { error => List(error.getMessage) }
     } match {
       case Right((ruleResource, lastModified)) => {
-        ruleProvisioner.updateRules(ruleResource, lastModified)
+        val ruleErrors = ruleProvisioner.updateRules(ruleResource, lastModified) match {
+          case Right(()) => Nil
+          case Left(errors) => errors.map(_.getMessage)
+        }
         val currentRules = matcherPool.getCurrentRules
         Ok(views.html.rules(
             sheetId,
@@ -39,7 +42,7 @@ class RulesController(
             matcherPool.getCurrentMatchers,
             Some(true),
             Some(currentRules.size),
-            Nil
+            ruleErrors
         ))
       }
       case Left(errors) => {
--- a/app/services/RuleProvisionerService.scala
+++ b/app/services/RuleProvisionerService.scala
@@ -28,27 +28,32 @@ class RuleProvisionerService(
   /**
     * Update the rules in our matcherPool, given a ruleResource.
     */
-  def updateRules(ruleResource: RuleResource, date: Date): Unit = {
+  def updateRules(ruleResource: RuleResource, date: Date): Either[List[Throwable], Unit] = {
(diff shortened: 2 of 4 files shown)


app/services/RuleProvisionerService.scala
```diff
@@ -25,27 +25,36 @@ class RuleProvisionerService(
   /**
     * Update the rules in our matcherPool, given a ruleResource.
     */
-  def updateRules(ruleResource: RuleResource, date: Date): Unit = {
+  def updateRules(ruleResource: RuleResource, date: Date): Either[List[Throwable], Unit] = {
     matcherPool.removeAllMatchers()
 
-    ruleResource.rules.groupBy(_.category).foreach {
+    addLTMatcherToPool(matcherPool, Nil, ruleResource.ltDefaultRuleIds)
```
SHession, line 31: Is it worth trying to capture errors from the default rules? At the moment they are discarded.

app/controllers/RulesController.scala
```diff
@@ -31,15 +31,18 @@ class RulesController(
       maybeRules.left.map { error => List(error.getMessage) }
     } match {
       case Right((ruleResource, lastModified)) => {
-        ruleProvisioner.updateRules(ruleResource, lastModified)
+        val ruleErrors = ruleProvisioner.updateRules(ruleResource, lastModified) match {
+          case Right(()) => Nil
+          case Left(errors) => errors.map(_.getMessage)
+        }
```
SHession, line 37: It would be nice to prevent ingesting rules if there are issues. This could possibly be managed by rolling back to the S3 version.

SHession (commented):
A couple of comments for your consideration.

app/services/RuleProvisionerService.scala
jonathonherbert: We don't pass any errors as a result of the enable op – but we probably should, for example if we're trying to enable something that doesn't exist! I'll add that logic.

app/controllers/RulesController.scala
jonathonherbert: Agreed. To keep this PR focused on making existing errors visible, I'd argued that'd belong in another PR.

SHession (approved):

PR #97, opened 2020-11-03
Add per-matcher timer logging

jonathonherbert:
## What does this change?

We could do with a slow log for Typerighter's matcher jobs. This PR adds per-matcher logging to the matcher pool, to help us identify slow-running jobs.

Because we send timing metrics, and because we don't know what 'slow' looks like yet, I haven't added a 'warn' status to slow logs – my hunch is that'll be better to do on the metrics/alerting side, where we can tune in response to system performance.

In addition to this, per @SHession's suggestion, I've added a Cloudwatch metric that's sent when our slow log is triggered in the `matcherPool` service. 

## How to test

The new unit tests for the `Timer` slow-log should pass.

Deploy to CODE. You should see timer logs for the `runMatchersForJob` task in our logging.

Tested on CODE, and I've add a graph (matcher jobs duration percentiles) to our Typerighter metrics radiator.

Screenshot 2020-11-03 at 16 51 22

When performance is problematic, we should be able to identify which matcher type and categories are causing the problem.

## Dev notes

The API of `Timer` has expanded to accommodate slow logging. I've added a callback to delegate an action to the caller when the slowlog is triggered, if supplied, to avoid coupling our Timer to opinions about what slow logging might mean beyond logging.

This does the job and avoids coupling, but I'm not convinced I've done it in a very ergonomic way – comments welcome!

### DIFF
--- a/app/matchers/LanguageToolMatcher.scala
+++ b/app/matchers/LanguageToolMatcher.scala
@@ -29,7 +29,7 @@ class LanguageToolFactory(
                            useLanguageModelRules: Boolean = false) extends Logging {
 
   def createInstance(ruleXMLs: List[LTRuleXML], defaultRuleIds: List[String] = Nil)(implicit ec: ExecutionContext): Either[List[Throwable], Matcher] = {
-    val language: Language = Languages.getLanguageForShortCode("en")
+    val language: Language = Languages.getLanguageForShortCode("en-GB")
     val cache: ResultCache = new ResultCache(10000)
     val userConfig: UserConfig = new UserConfig()
 
@@ -46,6 +46,7 @@ class LanguageToolFactory(
     // ... apart from those we'd explicitly like
     val errors = defaultRuleIds.foldLeft(List.empty[Throwable])((acc, ruleId) =>
       if (allRuleIds.contains(ruleId)) {
+        println(s"Enabling rule: ${ruleId}")
         instance.enableRule(ruleId)
         acc
       } else new Exception(s"Attempted to enable a default rule with id ${ruleId}, but the rule was not available on the instance") :: acc
@@ -71,6 +72,8 @@ class LanguageToolFactory(
     val maybeRuleErrors = getLTRulesFromXML(ltRuleXmls) match {
       case Success(rules) => rules.map { rule =>
         val ruleTry = Try {
+          println("Adding custom rule")
+          println(rule)
           instance.addRule(rule)
           instance.enableRule(rule.getId())
         }
@@ -127,7 +130,7 @@ class LanguageToolFactory(
     }
 
     // Temporarily hardcode language settings
-    val ruleXml = <rules lang="en">{rulesXml}</rules>
+    val ruleXml = <rules lang="en-GB">{rulesXml}</rules>
 
     val outputStream = new ByteArrayOutputStream()
     val writer = new OutputStreamWriter(outputStream)
--- a/app/matchers/LanguageToolMatcher.scala
+++ b/app/matchers/LanguageToolMatcher.scala
@@ -29,7 +29,7 @@ class LanguageToolFactory(
                            useLanguageModelRules: Boolean = false) extends Logging {
 
   def createInstance(ruleXMLs: List[LTRuleXML], defaultRuleIds: List[String] = Nil)(implicit ec: ExecutionContext): Either[List[Throwable], Matcher] = {
-    val language: Language = Languages.getLanguageForShortCode("en-GB")
+    val language: Language = Languages.getLanguageForShortCode("en")
     val cache: ResultCache = new ResultCache(10000)
     val userConfig: UserConfig = new UserConfig()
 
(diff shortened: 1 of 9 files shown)


SHession:
Is it worth logging the duration as a custom metric through CloudWatch providing us with the ability to alert if it gets too high?

test/scala/utils/TimerTest.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class TimerTest extends AnyFlatSpec with Matchers with IdiomaticMockito {
+  implicit val executionContext = ExecutionContext.Implicits.global
+
+  behavior of "time"
+
+  it should "not call `onSlowLog` when a task does not its slow log threshold" in {
```
SHession, line 19:
```suggestion
  it should "not call `onSlowLog` when a task does not exceed its slow log threshold" in {
```

test/scala/utils/TimerTest.scala
```diff
@@ -0,0 +1,62 @@
...
+    Timer.time(taskName = "task", slowLogThresholdMs = 100, onSlowLog = mockFunction){
+      Thread.sleep(200)
+    }
+
+    eventualDuration.toInt should be >= 200
+  }
+
+  behavior of "timeAsync"
+
+  it should "not call `onSlowLog` when a task does not its slow log threshold" in {
```
SHession, line 40:
```suggestion
  it should "not call `onSlowLog` when a task does not exceed its slow log threshold" in {
```

app/services/MatcherPool.scala
```diff
@@ -237,7 +241,11 @@ class MatcherPool(
         .check(MatcherRequest(blocksWithSkippedRangesRemoved))
         .map { matches => matches.map(_.mapMatchThroughBlocks(blocks)) }
 
-      futures.timeout(checkTimeoutDuration)(eventuallyCheck)
+      val taskName = s"MatcherPool.runMatchersForJob (type: ${matcher.getType}, categories: ${matcher.getCategories.map(_.name).mkString(", ")})"
```
SHession, line 244: Would it be useful to add the matcher type and category as markers?

SHession (commented):
Code changes look sensible and the tests pass locally. I've added a few comments for your consideration. Would this PR lead on to a separate PR to implement a CloudWatch alarm?

app/services/MatcherPool.scala
```diff
@@ -65,7 +67,9 @@ class MatcherPool(
   val maxQueuedJobs: Int = 1000,
   val checkStrategy: MatcherPool.CheckStrategy = MatcherPool.blockLevelCheckStrategy,
   val futures: Futures,
-  val checkTimeoutDuration: FiniteDuration = 5 seconds
+  val checkSlowLogDuration: FiniteDuration = 5 seconds,
+  val checkTimeoutDuration: FiniteDuration = 10 seconds,
+  val maybeCloudWatchClient: Option[CloudWatchClient] = None
```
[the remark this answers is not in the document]
jonathonherbert: Not sold on the idea that we'd always want to have one logging away – in tests for example, which are already busy with logs – but more concretely we'd need to know the stage to instantiate one, and I don't think that's something our `matcherPool` should care about.

app/services/MatcherPool.scala
jonathonherbert: Great point!

app/matchers/LanguageToolMatcher.scala
```diff
@@ -71,6 +72,8 @@ class LanguageToolFactory(
     val maybeRuleErrors = getLTRulesFromXML(ltRuleXmls) match {
       case Success(rules) => rules.map { rule =>
         val ruleTry = Try {
+          println("Adding custom rule")
+          println(rule)
```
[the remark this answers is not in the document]
jonathonherbert: nooo println no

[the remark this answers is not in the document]
SHession: There is another println on line 49 of the file.

app/services/MatcherPool.scala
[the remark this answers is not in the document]
SHession: I was thinking the `matcherPool` wouldn't need to care about stage as it wouldn't instantiate one itself, just be passed a non-optional version. The reason "dryRun" was added was to remove the need for it being optional so it may be more consistent. However I think both ways are fine and I'm happy to leave it.

app/matchers/LanguageToolMatcher.scala
```diff
@@ -46,7 +46,6 @@ class LanguageToolFactory(
     // ... apart from those we'd explicitly like
     val errors = defaultRuleIds.foldLeft(List.empty[Throwable])((acc, ruleId) =>
       if (allRuleIds.contains(ruleId)) {
-        instance.enableRule(ruleId)
```
SHession, line 49: Was this line intentionally removed?

jonathonherbert: Aw, jeez

SHession (approved):
LGTM!
Apart from the build failure: `sbt.librarymanagement.ResolveException: Error downloading biz.k11i:xgboost-predictor:0.3.0`

jonathonherbert:
Yep, looks like some transient dep resolution problems. I'll raise on DevX.

jonathonherbert:
... hmm, I did add a new dependency, possibly there's something wrong here

PR #98, opened 2020-11-05
Adds a `dryRun` param for local CloudWatch

SHession:
## What does this change?

This PR fixes an issue introduced by #94, causing issue with running Typerighter locally. It also prevents custom metrics from being logged when running in a local machine. This is achieved by adding a dry run parameter to the `CloudWatchClient` class.

## How to test

Run TR locally and confirm it works correctly. For other environments this should be a NO-OP.

### DIFF
--- a/app/utils/CloudWatchClient.scala
+++ b/app/utils/CloudWatchClient.scala
@@ -29,7 +29,7 @@ class CloudWatchClient(stage: String, dryRun: Boolean) extends Loggable {
 
     try {
       cloudWatchClient.map(_.putMetricData(request))
-      log.info(s"Published metric data")
+      log.info(s"Published $metric metric data. ${if (dryRun) "DRY RUN" else ""}")
     } catch {
       case e: Exception =>
         log.error(s"CloudWatch putMetricData exception message: ${e.getMessage}", e)
--- a/app/utils/CloudWatchClient.scala
+++ b/app/utils/CloudWatchClient.scala
@@ -11,11 +11,9 @@ object Metrics {
   val RulesNotFound = "RulesNotFound"
 }
 
-class CloudWatchClient(stage: String) extends Loggable {
+class CloudWatchClient(stage: String, dryRun: Boolean) extends Loggable {
 
-  private val builder = AmazonCloudWatchClientBuilder.defaultClient()
-
-  private lazy val cloudWatchClient = builder
+  private val cloudWatchClient = if(dryRun) None else Some(AmazonCloudWatchClientBuilder.defaultClient())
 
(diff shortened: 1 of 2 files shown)


app/utils/CloudWatchClient.scala
```diff
@@ -30,8 +28,8 @@ class CloudWatchClient(stage: String) extends Loggable {
     val request = new PutMetricDataRequest().withNamespace("Typerighter").withMetricData(datum)
 
     try {
-      val result = cloudWatchClient.putMetricData(request)
-      log.info(s"Published metric data: $result")
+      cloudWatchClient.map(_.putMetricData(request))
+      log.info(s"Published metric data")
     } catch {
       case e: Exception =>
         log.error(s"CloudWatch putMetricData exception message: ${e.getMessage}", e)
```
tjsilver, line 35: Were we going to log a warning here instead, or a message about what data would be sent?

jonathonherbert: I don't think this should be a warning, as the system is behaving as expected – but I definitely think we should log a) that this is a dry run, and b) what it is we were intending to post to CW.

app/utils/CloudWatchClient.scala
```diff
@@ -30,8 +28,8 @@ class CloudWatchClient(stage: String) extends Loggable {
     val request = new PutMetricDataRequest().withNamespace("Typerighter").withMetricData(datum)
 
     try {
-      val result = cloudWatchClient.putMetricData(request)
-      log.info(s"Published metric data: $result")
+      cloudWatchClient.map(_.putMetricData(request))
+      log.info(s"Published $metric metric data. ${if (dryRun) "DRY RUN" else ""}")
     } catch {
       case e: Exception =>
         log.error(s"CloudWatch putMetricData exception message: ${e.getMessage}", e)
```
tjsilver, line 35: Apologies if I am misunderstanding this - but won't this exception always be triggered on a dry run, because the cloudwatchClient will be None? How does the log.info for dry run get triggered if there is no cloudwatchClient for the putMetricData?

SHession: No exception will be thrown if `cloudWatchClient` is `None` because the `map` function will only map over a `Some` value. This means that we will only call `putMetricData` if we have a `cloudWatchClient`. The message should get logged if there is no `cloudWatchClient` or there is one and it's `put` successfully.

tjsilver: Got it! Thanks for explaining

tjsilver (approved):
👍

PR #99, opened 2020-11-06
Correctly reported checked categories

jonathonherbert:
## What does this change?

At the moment, we only report the categories that are checked when they produce matches – a major oversight, as our client depends on the Typerighter service to report the categories it has checked to invalidate matches that may have been affected by document changes.

To reproduce the problem:
- Check the text: `We an hope to address it.` We should get a match for `an`.
- and a `c` to correct.
- Check again. The match will remain, incorrectly.

This PR ensures that we pass back the categories that the `matcherPool` has checked in our API response.

## How to test

Attempt the reproduction above with this change. The match in question should disappear on a re-check.

The unit test added for this issue should pass.

Changes to the document that resolve matches but do not directly touch them should be removed on subsequent checks.

### DIFF
--- a/test/scala/services/MatcherPoolTest.scala
+++ b/test/scala/services/MatcherPoolTest.scala
@@ -164,7 +164,7 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
 
   behavior of "check"
 
-  it should "return a list of MatcherResponses" in {
+  it should "return a list of matches and categoryIds checked" in {
     val matchers = getMatchers(1)
     matchers.head.completeWith(responses)
 
@@ -172,6 +172,7 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
     val futureResult = pool.check(getCheck(text = "Example text"))
     futureResult.map { case (categoryIds, matches) =>
       matches shouldBe responses
+      categoryIds should matchTo(Set(getCategory(0).id))
     }
   }
 
@@ -258,7 +259,7 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
     eventualResult
   }
 
-  it should "correctly check multiple categories for a single job" in {
+  it should "correctly check multiple categories for a single job, and report them" in {
     val matchers = getMatchers(2)
     val firstMatch = getResponses(List((0, 5, "test-response")), 0)
     val secondMatch = getResponses(List((6, 10, "test-response")), 1)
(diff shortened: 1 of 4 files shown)


test/scala/services/MatcherPoolTest.scala
```diff
@@ -352,9 +364,9 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
     )
     val futureResult = pool.check(check)
 
-    futureResult.map { results =>
-      val resultRanges = results.map { result => (result.fromPos, result.toPos) }
-      resultRanges shouldBe List((1, 1), (5, 5))
+    futureResult.map { case (categoryIds, matches) =>
```
jonathonherbert, line 367: A few renames here to better reflect the data being passed.

test/scala/services/MatcherPoolTest.scala
```diff
@@ -268,9 +268,21 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
...
-    futureResult.map { result =>
-      result.contains(firstMatch.head) shouldBe true
-      result.contains(secondMatch.head) shouldBe true
+    futureResult.map { case (categoryIds, matches) =>
+      matches.contains(firstMatch.head) shouldBe true
+      matches.contains(secondMatch.head) shouldBe true
+    }
+  }
+
+    it should "report the categories that are checked, even if no matches are found" in {
```
SHession, line 277:
```suggestion
  it should "report the categories that are checked, even if no matches are found" in {
```

test/scala/services/MatcherPoolTest.scala
```diff
@@ -282,9 +294,9 @@ class MatcherPoolTest extends AsyncFlatSpec with Matchers {
     val secondMatch = getResponses(List((0, 5, "test-response-2")), 1)
     matchers(0).completeWith(firstMatch)
     matchers(1).completeWith(secondMatch)
-    futureResult.map { result =>
-      result.size should matchTo(1)
-      result should matchTo(secondMatch)
+    futureResult.map { case (categoryIds, matches) =>
+      matches.size should matchTo(1)
```
SHession, line 298: Is it worth adding some category checks to the existing tests?

SHession (commented):
Code changes look good but I've been unable to test due to the CloudWatch issues. Could you rebase master to allow local testing?

jonathonherbert:
I've just rebased, and added category checks for single/multiple cats 👍

SHession (approved):
I've now tested locally, works as described!

PR #100, opened 2020-11-17
Add decision record for management service

jonathonherbert:
## What does this change?
Adds a decision record for the rule management service.

## How to test

Read the doc. Anything to add? Challenge? Change?

We've got consensus on the technical foundations of the service.


decision-records/001-management-service.md
```diff
@@ -0,0 +1,143 @@
...
+
+-   Browsing, sorting, and filtering rules (RO)
+
+-   CRUD for rules, including the ability to evaluate rules in a draft context, before applying them to the published corpus (RO)
+
+-   Understanding the history of a rule, including seeing when and what was changed by whom (RO, RU)
+
+-   Adding feedback to rules based on match performance (RU)
+
+See the [trello board](https://trello.com/b/PlsiUqYq/typerighter) for a comprehensive list.
```
tjsilver, line 19: Maybe we could copy the relevant list from Trello to another doc in this repo, so that it persists, as the board is likely to change when we do another wave of work.

jonathonherbert: Great idea 👍

SHession (approved):
This looks great, agree with @tjsilver's above comment. This file would likely be better placed in the `docs` folder rather than a new one on the root.

jonathonherbert:
Added user stories 👍

PR #102, opened 2020-11-19
Generate (more) correct suggestions for case-insensitive regexes

jonathonherbert:
## What does this change?

At the moment, when we apply suggestions from regular expressions, there's sometimes an unexpected side effect: we overwrite casing in the matched text.

For example, with regex `(?i)\\bmedia?eval` (note the `(?i)` flag, which means it's case insensitive), we always suggest the word `medieval`. 

This causes problems when words begin sentences, e.g. `end of sentence. Medieval` will produce a match suggesting `medieval`.

This PR preserves case where possible:
- When casing is ignored, if the suggestion matches the matched text, replace the suggestion with the matched text
- When casing is ignored, if the suggestion doesn't match the whole match text, but the first char does, preserve the case of the first char to ensure we don't accidentally correct start-of-sentence casing
- Otherwise, continue as normal

The reason we preserve the suggestion on a perfect caseless match, rather than just stripping it, is to preserve the match's 'mark as correct' behaviour.

## How to test

The unit tests should pass.

E.g. the sentence `End of sentence. Mediaeval` should offer a correction to `Medieval`. Before, it would offer `medieval`.

Fewer complaints about mistakes w/ casing in suggestions.

This is still not perfect – see the 'first character' kludge above! – but the rules to which this apply are probably better addressed in the long run with a dictionary.

### DIFF
--- a/checker/app/matchers/RegexMatcher.scala
+++ b/checker/app/matchers/RegexMatcher.scala
@@ -1,11 +1,14 @@
 package matchers
 
-import model.{RegexRule, RuleMatch, Category}
-import services.MatcherRequest
-import utils.{Matcher, MatcherCompanion, RuleMatchHelpers}
+import java.util.regex.Pattern
 
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext
+import scala.util.matching.Regex
+import model.{RegexRule, RuleMatch}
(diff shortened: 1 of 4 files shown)


tjsilver (approved):
👍 Tested locally and all tests passed.

PR #103, opened 2020-11-19
Add rule manager subproject

SHession:
## What does this change?

This PR adds a Rule Manager sub project to the Typerighter repo. It provides us with a starting point for developing the Rule Manager functionality. This PR includes the Play setup but doesn't go as far as to configure the deploy which will be covered in a separate PR. 

## How to test

From the perspective of the existing checker functionality, this should be a no-op. In terms of the Rule Manager it should be possible to run locally on [typerighter.local.dev-gutools.co.uk](https://typerighter.local.dev-gutools.co.uk/), this will require nginx to be setup again to account for the change of ports.

We are able to proceed with the rule management functionality.


rule-manager/conf/logback.xml
```diff
@@ -0,0 +1,41 @@
+<!-- https://www.playframework.com/documentation/latest/SettingsLogger -->
+<configuration>
```
jonathonherbert, line 2: I wonder if there's a way for us to share logback configs (esp. WRT kinesis) – we can explore in another PR!

SHession: Definitely a good idea. As you say, I wouldn't want to block the progress of this prerequisite work, so one for another PR.

jonathonherbert:
Be good to have the script start everything – for example, Grid has a `runAll` command.

Having said that, when I try to define an alias, I get port conflicts – e.g. addCommandAlias("runAll", "all ruleManager/run checker/run") invariably results in a port conflict, possibly because settings in one app are leaking into another. Perhaps I'm doing it wrong?

SHession:
> Be good to have the script start everything – for example, Grid has a `runAll` command.
 Agreed, I tried this locally and had the same problems. I tried for sometime to find a workaround but I haven't be able to get it working. I don't think this is a blocker for this PR but will definitely be useful when we have both subprojects up and running.

jonathonherbert (approved):
👍 Cracking start. I'd probably get rid of the conf/messages and main.js/main.css, as AFAICS we're unlikely to   use 'em.

SHession:
> 👍 Cracking start. I'd probably get rid of the conf/messages and main.js/main.css, as AFAICS we're unlikely to use 'em.

I left those in partly due to the fact git needs a file to version a directory. Happy to remove them if you feel strongly.

PR #104, opened 2020-11-19
Use corrected suggestions when computing markAsCorrect

jonathonherbert:
## What does this change?

#102 transforms suggestions, but still uses the untransformed suggestion to calculate `markAsCorrect`, leading to misleading rules – the first match in the following should be a 'green' rule.

Screenshot 2020-11-19 at 17 36 39

This PR corrects that oversight.

## How to test

The unit test should pass.

Does the copy in the screenshot now produce the correct colour?

### DIFF
--- a/checker/app/controllers/ApiController.scala
+++ b/checker/app/controllers/ApiController.scala
@@ -42,7 +42,10 @@ class ApiController(
           }
         } recover {
         case e: Exception =>
-          InternalServerError(Json.obj("error" -> e.getMessage))
+          log.error(checkMarkers, "Error in /check", e)
+          InternalServerError(Json.obj(
+            "error" -> e.getMessage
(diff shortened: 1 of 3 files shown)


checker/app/controllers/ApiController.scala
```diff
@@ -42,7 +42,10 @@ class ApiController(
           }
         } recover {
         case e: Exception =>
-          InternalServerError(Json.obj("error" -> e.getMessage))
+          log.error(checkMarkers, "Error in /check", e)
```
jonathonherbert, line 45: This is to ensure that any errors we receive are caught in our logs – ATM, we're just swallowing these errors from the POV of our logging.

SHession (approved):
LGTM

PR #108, opened 2020-11-23
RegexMatcher: address capitalisation problems at sentence starts

jonathonherbert:
## What does this change?

At the moment, when we apply suggestions from regular expressions, there's sometimes an unexpected side effect: we overwrite casing in the matched text.

For example, with regex `(?i)\\bmedia?eval` (note the `(?i)` flag, which means it's case insensitive), we always suggest the word `medieval`. 

This causes problems when words begin sentences, e.g. `end of sentence. Medieval` will produce a match suggesting `medieval`.

This PR preserves case where possible, by detecting sentence starts. When a regex applies to a sentence start, and the starting characters of the suggestion and the match whilst ignoring case, we keep the casing of the match.

So given the regex `(?i)\bmedia?eval` with the replacement `medieval` where square brackets denote a match:

- `... [mediavel] castles` will produce the suggestion `medieval`
- `end of sentence. [Medieaval] castles` will produce the suggestion `Medieval`
- `end of sentence. [medieaval] castles` will produce the suggestion `Medieval`

The reason we preserve the suggestion on a perfect caseless match, rather than just stripping it, is to preserve the match's 'mark as correct' behaviour.

## How to test

The unit tests should pass.

E.g. the sentence `End of sentence. Mediaeval` should offer a correction to `Medieval`. Before, it would offer `medieval`.

Fewer complaints about mistakes w/ casing in suggestions.

This is probably still not perfect, but the rules to which this apply are probably better addressed in the long run with a dictionary. Wonky edge cases gratefully received.

### DIFF
--- a/checker/app/model/BaseRule.scala
+++ b/checker/app/model/BaseRule.scala
@@ -83,14 +83,13 @@ case class RegexRule(
     // A kludge to get around start-of-sentence casing. If the suggestion doesn't
     // match the whole matchedText, but does match the first character, preserve that
     // casing in the suggestion. This is to ensure that e.g. a case-insensitive suggestion
-    // to replace 'end of sentence. [Mediavel]' with 'medieval' does not incorrectly replace
+    // to replace e.g. 'end of sentence. [Mediavel]' with 'medieval' does not incorrectly replace
     // the uppercase 'M'.
     //
     // These sorts of rules are better off as dictionary matches, which we hope to add soon.
-    case TextSuggestion(text)
-      if isStartOfSentence && text.charAt(0).toLower == matchedText.charAt(0).toLower => {
-        TextSuggestion(text = matchedText.charAt(0) + text.slice(1, text.length))
-      }
+    case TextSuggestion(text) if isStartOfSentence => {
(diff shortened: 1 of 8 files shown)


checker/app/model/BaseRule.scala
```diff
@@ -69,6 +73,25 @@ case class RegexRule(
       matcherType = RegexMatcher.getType
     )
   }
+
+  /**
+    * If the first character of the suggestion is identical to the first character
+    * of the matched text case, preserve the original casing. Used when our match covers
+    * the start of a sentence to ensure we don't accidentally lowercase sentence starts.
+    */
+  private def maybePreserveMatchCase(isStartOfSentence: Boolean, suggestion: Suggestion, matchedText: String): Suggestion = suggestion match {
```
aug24, line 82: Any reason this isn't a method on the Suggestion class?

aug24 (changes requested):

checker/app/model/BaseRule.scala
jonathonherbert: Yeah, that's much better, ta.

aug24 (approved):

PR #109, opened 2020-11-24
Raise timeout limit

jonathonherbert:
## What does this change?

Some of our requests are timing out, but the service is still healthy. Timing out should be a last resort when something's gone wrong, so let's raise the limit.

This is not a substitute for a timely service – it looks like #108 has slowed us up somewhat, and we should investigate.

## How to test

Test with a large document in CODE. You should not see requests time out in 10s.

### DIFF
--- a/checker/app/services/MatcherPool.scala
+++ b/checker/app/services/MatcherPool.scala
@@ -68,7 +68,7 @@ class MatcherPool(
   val checkStrategy: MatcherPool.CheckStrategy = MatcherPool.blockLevelCheckStrategy,
   val futures: Futures,
   val checkSlowLogDuration: FiniteDuration = 5 seconds,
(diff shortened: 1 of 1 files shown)


SHession (approved):
This seems sensible, especially as we have the slow log warning. Do we need an alarm for the slow logs crossing a certain threshold?

jonathonherbert:
> This seems sensible, especially as we have the slow log warning. Do we need an alarm for the slow logs crossing a certain threshold?

It's a good idea, I think – the project is active now so we've got 👀  on metrics, but once it's ticking along in the background these sorts of changes could drop unnoticed.

It'd also be nice to automate some load testing in CI so we have a rough idea of performance changes before we merge. We've got Gatling tests in the project that we could leverage as one way forward.

PR #112, opened 2020-11-25
Ignore metals/bloop

sihil:
## What does this change?
Remove metals config and modify .gitignore to avoid it being checked in again.

No more accidental metals for developers that are not yet drinking the koolaid.


jonathonherbert (approved):
🙏

PR #113, opened 2020-12-08
Create rule manager client

tjsilver:
## What does this change?
Adds a React/Typescript client to the rule manager app.

## How to test

In the root of the branch, run `sbt "ruleManager/run"` and navigate to https://typerighter.local.dev-gutools.co.uk/ where you should see the message "Hello world! This is the rule manager client!".

Users will be (eventually) able to manage Typerighter rules through a website rather than having to manually edit a spreadsheet.


tjsilver:
[Management: create 'hello world' React frontend for service](https://trello.com/c/QpsEoAMW/519-management-create-hello-world-react-frontend-for-service)

checker/app/views/manager.scala.html
```diff
@@ -0,0 +1,4 @@
+@main(title = "Rule manager", pageName = Some("Rule manager"), breadcrumbsList = List("Manager")) {
```
jonathonherbert, line 1: Is this used?

rule-manager/rule-manager-client/dist/main.js
```diff
@@ -0,0 +1,12 @@
+/*
```
jonathonherbert, line 1: Presume this isn't necessary to store in VCS?

checker/app/views/manager.scala.html
tjsilver: No, don't think so - will remove

jonathonherbert (approved):
Tested locally, works as expected! Looks like the relevant parts of the build are created in prod, but we'll only really be able to test this once we've a deployable service.

PR #114, opened 2020-12-09
Add a postgres database to the rule management service

jonathonherbert:
## What does this change?

Adds a DB integration using [scalikejdbc](http://scalikejdbc.org/). Uses the db ops trait pattern found in [facia-tool](https://github.com/guardian/facia-tool/blob/main/app/services/editions/db/EditionsDB.scala).

In light of the issues we've had testing in different CI environments in the Grid (e.g. GHA), we've opted for a docker-compose call within a script for local and CI in place of a Dockerkit based solution. This also means that both local and test rely on the same set of definitions for their environment, but v. happy to consider other solutions if anyone has other suggestions/opinions.

## How to test

The healthcheck should pass, and the tests should pass locally and in CI – they're all now dependent on an active database connection.

We have a database implementation that works in our local and CI environments, and it's easy to test against.

Infra work to follow in a separate PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,10 @@ val commonSettings = Seq(
     )
   },
   libraryDependencies ++= Seq(
-    "net.logstash.logback" % "logstash-logback-encoder" % "6.0"
+    "net.logstash.logback" % "logstash-logback-encoder" % "6.0",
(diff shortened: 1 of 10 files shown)


SHession (approved):
Paired on these changes so happy they work correctly. There are some conflicts with `Main` which need to be resolved before this can be merged.

PR #115, opened 2020-12-09
Consolidate application code in a single folder

jonathonherbert:
## What does this change?

This PR consolidates application code in a single folder. 

In this way, it's clear to devs as they look at the project where to go for app code (global or per-project), and where to go for infra, configuration etc.

Done as I realised that adding whilst adding a 'common' folder it wasn't clear it was going to be code related.

## How to test

This is a no-op – the app should run locally, test, and deploy as it did before.

This PR has many changes, but they should consist of either path changes and/or file moves.

Arguably slightly less confusion when navigating repo? 😀


SHession (approved):
This looks like a sensible change and this works correctly locally.
The `public` folder in the apps directory doesn't look correct to me.

jonathonherbert:
Agreed. Turns out we can scope them locally, so I've removed that folder 👍

PR #116, opened 2020-12-10
Add project for common dependencies

jonathonherbert:
## What does this change?

Adds a folder to share common dependencies across projects, and moves the code we expect to share with the rule manager service – Elk logging, shared logging interfaces, and panda authentication.

## How to test

This is a no-op, everything should work as before.

Does everything added to the shared folder belong? Might be worth checking out locally and having a look around.


jennygrahamjones (approved):
Looks good locally to me 👍

PR #118, opened 2020-12-11
Add edit permissions

jonathonherbert:
## What does this change?

Adds edit permissions to a temporary endpoint, `/editPermissionCheck`, via our [permissions](https://github.com/guardian/permissions) service. Production use will depend on an [open PR](https://github.com/guardian/permissions/pull/129) in that service. (edit: this is now merged.)

## How to test

Locally, hit the `/editPermissionCheck` endpoint whilst running the rule manager service (`sbt ruleManager / run`). You should be denied permission.

Add CODE permissions for the `manage_rules` permission, and wait a few minutes for them to be published and for your local perm cache to clear (or restart the app!). You should be granted permission.

We can use permissions as we please in the project.

## Dev notes

I've used a config DTO as per [e.g. Grid](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala) to easily pass along configuration settings without increasing the arity of common class constructors – I hope this pattern makes sense!

### DIFF
--- a/apps/checker/app/AppComponents.scala
+++ b/apps/checker/app/AppComponents.scala
@@ -1,4 +1,4 @@
-import java.io.File
+
 
 import scala.concurrent.Future
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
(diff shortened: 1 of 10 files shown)


jennygrahamjones (approved):
Tested locally and behaves as expected 👍 👍

PR #120, opened 2020-12-14
Create .nvmrc

jennygrahamjones:
When running the setup script for the first time in a while, I encountered an error due to using an incorrect node version.

I'm not sure what the project stance on using .nvmrc files. Should I be detecting the node version and changing it in a different way? 🤔


jonathonherbert (approved):
Yeah, it's a great question that I haven't taken a good look at yet, but it's important we solve. I think @AWare might recommend us using yarn for this purpose, as IIRC it enforces the `engines` prop in `package.json`.

In the mean time, no objection to an `.nvmrc`.

PR #122, opened 2020-12-18
Setup rule manager infrastructure

SHession:
## What does this change?

This PR sets up the necessary infrastructure for the new Rule Manager service. This includes a Web and DB stack.

The CDK is using the [@guardian/cdk ](https://github.com/guardian/cdk) package as a guinea pig project and also to ensure we stay aligned with future changes to how we handle our infrastrucutre. 

This change primarily sets up a pipeline to create the CloudFormation during CI and deploy them via riffRaff.

I'm happy to provide any additional detail to this description which would be useful. 

## How to test

This PR can be tested by ensuring that it run correctly in both the local and CODE environments. 
- For local this change should be a no-op and the rule-manager should continue to work as normal
- For CODE it should be possible to deploy the stacks and app and see the ELB healthchecks pass successfully.

It is possible to deploy a working version of the Rule Manager and we have a CI and CD pipeline which enables us to iterate quickly. We are also easily able to make any necessary changes to our infrastructure through use of CDK.

During the development of this PR there was an issue found with the RDS password. Namely, the password parameter versions need to be inline with each other. Otherwise there will be issues deploying. This is likely to make changing the password frustrating in the future.  We might consider using some mapping for the password version number across environments, but the ideal solution would be to always use the latest (not currently possible with the `resolve` syntax).


apps/rule-manager/cfn/cfn.yaml
```diff
@@ -0,0 +1,211 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: Typerighter rule-management tool
+Parameters:
```
jonathonherbert, line 3: Do we want to check our generated YAML in? Or is this something else?

cdk/README.md
```diff
@@ -0,0 +1,15 @@
+# Welcome to your CDK TypeScript project!
+
```
jonathonherbert, line 2: Might be worth adding some detail about how the stacks are set up. V. happy to contribute this change on Mon, or pair.

cdk/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "aws-cdk": "~1.74.0",
+    "jest": "^26.4.2",
+    "ts-jest": "^26.2.0",
+    "ts-node": "^9.0.0",
+    "typescript": "~3.9.7"
+  },
+  "dependencies": {
+    "@aws-cdk/aws-sns": "~1.74.0",
+    "@aws-cdk/aws-sns-subscriptions": "~1.74.0",
+    "@aws-cdk/aws-sqs": "~1.74.0",
```
jonathonherbert, line 27: Do we need sqs or sns deps here?

cdk/test/rule-manager.test.ts
```diff
@@ -0,0 +1,21 @@
+import { expect as expectCDK, haveResource } from '@aws-cdk/assert';
+import * as cdk from '@aws-cdk/core';
+import * as Cdk from '../lib/rule-manager/rule-manager';
+
+test('SQS Queue Created', () => {
```
jonathonherbert, line 5: Presume these won't pass at present, we could replace these with a snapshot test re: @akash1810's recommendation, again v. happy to take a look Monday

jonathonherbert:
This is looking great 👍  a few things to pick up next week.

cdk/lib/rule-manager/rule-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+mkdir /etc/gu
+
+cat > /etc/gu/typerighter-rule-manager.conf <<-'EOF'
+    include "application"
+EOF
+
+aws --quiet --region ${this.region} s3 cp s3://composer-dist/${this.stack}/${this.stage}/typerighter-rule-manager/typerighter-rule-manager.deb /tmp/package.deb
+dpkg -i /tmp/package.deb`;
+
+    new GuAutoScalingGroup(this, "AutoscalingGroup", {
```
akash1810, line 161: A lot of boilerplate! `@guardian/cdk` should do this for free, hopefully we'll address this soon!

cdk/test/rule-manager.test.ts
```diff
@@ -0,0 +1,21 @@
+import { expect as expectCDK, haveResource } from '@aws-cdk/assert';
```
akash1810, line 1: Is this file needed? It doesn't test any part of stacks being created.

akash1810 (commented):
Exciting!

apps/rule-manager/cfn/cfn.yaml
SHession: This was leftover from when I was implementing the cfn as a yaml file. It's no longer needed and I will remove it.

cdk/package.json
SHession: Not needed I will remove.

cdk/test/rule-manager.test.ts
SHession: I will add some detail into here.

apps/rule-manager/cfn/cfn.yaml
```diff
@@ -0,0 +1,211 @@
+AWSTemplateFormatVersion: 2010-09-09
```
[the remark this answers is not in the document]
SHession: I have removed this as it's no longer needed. It was added before we decided to use the CDK.

cdk/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.ts
+!*.d.ts
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
```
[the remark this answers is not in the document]
SHession: Good spot on this, I have removed it as it's unnecessary.

cdk/.nvmrc
```diff
@@ -0,0 +1 @@
+v15.4.0
```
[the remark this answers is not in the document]
SHession: Yeah you're right, not sure how that happened. I've changed it to 14.15.3

cdk/test/rule-manager.test.ts
SHession: This is from the CDK template. I would recommend removing them and then implementing our testing in a separate PR. Happy to include it in this PR if you feel strongly.

jonathonherbert (approved):
Looks good 👍  be interested to know if @akash1810 has anything else to add, but otherwise, ✨

akash1810 (approved):
Looks good 👍🏽 . Well, it's too verbose, but that's a failing on the cdk library which we'll address soon.

PR #124, opened 2021-01-04
Spruce readme

jonathonherbert:
## What does this change?

A spruce of the readme to better describe the project – and its relation to the open source projects it depends upon.

## How to test

Give it a read! Anything to change/add?


SHession (approved):
This is a great improvement, some gifs or images might also be a nice addition.

jonathonherbert:
Great point @SHession, added – the checker isn't too exciting but at least gives an example. Roll on rule management 👍

PR #126, opened 2021-01-22
Convert typerighter to use graviton instances

sihil:
## What does this change?
Switch from t3 to t4g class instances which are cheaper but also Graviton (so we need to update the AMI as part of the conversion).

## How to test
Deploy. Should change the AMI and instance type.

Reduction in cost. Possible improvement in performance.

The risk is deemed low. A number of other services have already been switched.


sihil:
This has been successfully deployed to CODE.

SHession (approved):
This change was demoed in the Tech Time ands deploys successfully to CODE. It should increase speed and reduce costs!

PR #128, opened 2021-01-27
Update gu-cdk

jonathonherbert:
## What does this change?

Updates us to the latest version of `gu-cdk`, and makes use of its new, boilerplate busting features.

Work in progress.

## How to test

The snapshot tests (TBA!) should pass. The generated cloudformation should work correctly in CODE & PROD.


jonathonherbert:
Closed in favour of #131.

PR #132, opened 2021-03-25
Changes RDS postgress db version to Major 11

SHession:
## What does this change?

This PR changes the Postgres engine version to "11" from "11.8". This was necessary due to infra drift, causing failing deploys. The RDS instances are currently on "11.10" and were rejecting deploys due to the downgrade in version. Changing the version to major 11 allows the deploy to continue as normal. 

Please note, there is no 11.10 currently available in the CDK. The highest available version is [11.9](https://docs.aws.amazon.com/cdk/api/latest/typescript/api/aws-rds/postgresengineversion.html#aws_rds_PostgresEngineVersion_VER_11_9)

## How to test

Are we able to deploy as normal? We no longer see failed deploys like [this one](https://riffraff.gutools.co.uk/deployment/view/39372a89-acfe-4d71-a8b0-a53022bc45c6).


tjsilver (approved):
👍

PR #134, opened 2021-04-15
Account for quotes at start of sentences when detecting sentence start

jonathonherbert:
## What does this change?

Quotes are represented by their own tokens in CoreNLP. Previous to this change, we selected the first token at the sentence boundary as the sentence word start. If this token happens to be a word, our uppercase rules succeed – but if it's a quote, we apply uppercase rules to the quote token instead of the token representing the first word, and get silly suggestions:

`... applies to holiday accomodation. \"Cafes\", bars ...`

`Cafe` -> `cafe`

This patch scans forward for the first word token it sees, finding the first word proper. It also accounts for parentheses, square brackets and dashes of various kinds, which had a similar effect. Other chars should be easy to account for if we think of them.

## How to test

The automated tests should look comprehensive, and pass.

The following text should produce green matches for the words 'Cafe', indicating that we've skipped the first, non-word token in the sentence:

```
“Cafes will open soon”

‘Cafes will open soon’

`Cafes will open soon`

(Cafes will open soon.)

[Cafes will open soon]

Cafes will open soon
```

### DIFF
--- a/apps/checker/test/scala/services/SentenceHelperTest.scala
+++ b/apps/checker/test/scala/services/SentenceHelperTest.scala
@@ -18,30 +18,24 @@ class SentenceHelperTest extends AsyncFlatSpec with Matchers {
     ))
   }
 
-  it should "ignore quotes when finding sentence starts – double quotes" in {
-    val sentenceTokenizer = new SentenceHelpers()
-    val firstWordsInSentences = sentenceTokenizer.getFirstWordsInSentences("Allowed to have up to 15 people in their home per day, and this rule applies to holiday accomodation. \"Cafes\", bars, and restaurants will be able to seat 100 indoors and 200 outdoors, within the density limit")
-    firstWordsInSentences should matchTo(List(
-      WordInSentence("Allowed to have up to 15 people in their home per day, and this rule applies to holiday accomodation.", "Allowed", TextRange(0, 7)),
-      WordInSentence("\"Cafes\", bars, and restaurants will be able to seat 100 indoors and 200 outdoors, within the density limit", "Cafes", TextRange(103, 108)),
-    ))
-  }
-
-  it should "ignore quotes when finding sentence starts – single quotes" in {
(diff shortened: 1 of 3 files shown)


paperboyo:
Maybe a dash/+space, too if possible?
`end of preceding. – Cafes are open, he said.`
? (it’s an en dash, but could also be a hyphen if the cleaner/dash rule wasn’t yet used)

jonathonherbert:
Great point, those cases fail our tests and are easily accounted for.

apps/checker/test/scala/services/SentenceHelperTest.scala
```diff
@@ -17,4 +17,28 @@ class SentenceHelperTest extends AsyncFlatSpec with Matchers {
...
+    ("{", "}"),
+    ("[", "]"),
+    ("(", ")"),
+    ("“", "”"),
+    ("‘", "’)"),
+    ("-", ""),
+    ("–", ""),
+    ("—", "")
+  ).foreach {
+    case (leftNonWordChar, rightNonWordChar) =>
```
SHession, line 37: Out of interest, why do we need the "rightNonWordChar" for this test?

SHession (approved):
LGTM!

apps/checker/test/scala/services/SentenceHelperTest.scala
jonathonherbert: Mm, good point – we don't! It just felt a bit silly to provide something which didn't look like proper text – for example, w/ open parens. Arguably this is just confusing for users, happy to get rid – or provide a comment to explain.

SHession: I would lean towards removing them, but it's not a big deal if you want to leave it.

jonathonherbert: Yeah, now you've mentioned it I agree. I'll do that.

PR #136, opened 2021-04-22
chore: Add snapshot tests for CDK stacks

akash1810:
## What does this change?

Snapshot tests make it easier to see what changes are made when the version of @guardian/cdk is changed.

This is a no-op.

## How to test

Observe CI?

## How can we measure success?

We have confidence in future PRs that update @guardian/cdk as we can see the exact changes they make.

## Have we considered potential risks?

n/a

## Images

n/a

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

PR #137, opened 2021-04-22
chore(deps): Update @guardian/cdk to v12.0.0

akash1810:
Requires #136.

## What does this change?

BREAKING CHANGE: Some security groups will be replaced

Releasing this change will require some careful orchestration during deployment.

Security group replacements cannot be done in place. Let us have an ASG with two machines, Q and W. Q and W are running and have been added to the security group E.

This PR will replace E with a new security group F. This cannot happen in place - AWS will delete E and then create F. However, as E is in use, it cannot be deleted and the CloudFormation event log will report an error that looks like this:

```
resource sg-ABC has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation;)
```

Thankfully, CloudFormation doesn't exit immediately on this error, instead it retries. We can use this to our advantage and perform an app only deploy to replace Q and W and thus E isn't used and can be removed and the CFN update completes. All w/out downtime.

That is the careful orchestration is:
  - Deploy CFN only
  - Wait for the `DependencyViolation` error
  - Perform an app only deploy
  - Observe the CFN update complete

## How to test

???

Keeping up to date.

Releasing this change will require some careful orchestration during deployment.

## Images

n/a

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/cdk/bin/rule-manager.ts
+++ b/cdk/bin/rule-manager.ts
@@ -5,10 +5,8 @@ import { RuleManagerDB } from "../lib/rule-manager/rule-manager-db";
 
 const app = new App();
 new RuleManager(app, "rule-manager", {
-  app: "typerighter-rule-manager",
   stack: "flexible",
 });
 new RuleManagerDB(app, "rule-manager-db", {
-  app: "typerighter-rule-manager",
   stack: "flexible",
 });
--- a/cdk/lib/rule-manager/__snapshots__/rule-manager-db.test.ts.snap
+++ b/cdk/lib/rule-manager/__snapshots__/rule-manager-db.test.ts.snap
@@ -33,7 +33,7 @@ Object {
     },
   },
   "Resources": Object {
-    "DBSecurityGroup": Object {
+    "DBSecurityGroupTyperighterrulemanagerE0229216": Object {
       "Properties": Object {
         "GroupDescription": "DB security group servers",
         "SecurityGroupEgress": Array [
@@ -50,7 +50,7 @@ Object {
           },
           Object {
             "Key": "gu:cdk:version",
-            "Value": "0.39.2",
+            "Value": "12.0.0",
           },
           Object {
             "Key": "Stack",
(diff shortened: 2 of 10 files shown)


akash1810:
After testing this on CODE, we've witnessed the change set attempting a load balancer replacement even though the change set is this:

```json5
  {
    "resourceChange": {
      "logicalResourceId": "PublicLoadBalancerXXXX",
      "action": "Remove",
      "physicalResourceId": "arn:aws:elasticloadbalancing:eu-west-1:XXXX:loadbalancer/app/Typer-Publi-XXXX",
      "resourceType": "AWS::ElasticLoadBalancingV2::LoadBalancer",
      "replacement": null,
      "moduleInfo": null,
      "details": [],
      "changeSetId": null,
      "scope": []
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
  {
    "resourceChange": {
      "logicalResourceId": "PublicLoadBalancerTyperighterrulemanagerXXXX",
      "action": "Add",
      "physicalResourceId": null,
      "resourceType": "AWS::ElasticLoadBalancingV2::LoadBalancer",
      "replacement": null,
      "moduleInfo": null,
      "details": [],
      "changeSetId": null,
      "scope": []
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
```

Switching this to draft whilst we understand what's going on here!

Apologies for creating a mess on CODE!

PR #138, opened 2021-04-22
chore: Use @guardian/cdk mock for tracking tag

akash1810:
Requires #137.

## What does this change?

The value of the tracking tag maps to the version of @guardian/cdk used.

This means bumping the version in package.json isn't merely enough as the snapshots will be out of date.
That is, PRs raised by Dependabot to automatically update @guardian/cdk will always fail.

By using the mock, the value of the tracking tag becomes static and thus Dependabot PRs become useful.

See:
  - https://github.com/guardian/cdk/blob/b025f27d562c8897fa0b831612609a7e6b51b057/docs/001-general-usage.md#-mocking-gucdkversion

## How to test

See CI?

We unlock Dependabot PRs 🎉 .

n/a

## Images

n/a

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/cdk/.gitignore
+++ b/cdk/.gitignore
@@ -1,5 +1,6 @@
 *.js
 !jest.config.js
+!jest.setup.js
 *.d.ts
 node_modules
 
--- a/cdk/jest.config.js
+++ b/cdk/jest.config.js
@@ -3,5 +3,6 @@ module.exports = {
   testMatch: ['**/*.test.ts'],
   transform: {
     '^.+\\.tsx?$': 'ts-jest'
-  }
+  },
+  setupFilesAfterEnv: ['<rootDir>/jest.setup.js']
 };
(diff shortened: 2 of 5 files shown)


jonathonherbert (approved):

jonathonherbert (approved):
Seems sensible, and the snapshots look stable as a result 👍

PR #140, opened 2021-04-28
Remove bintray resolver

jonathonherbert:
## What does this change?

Removes the Bintray resolver.

## Details

To accomplish these we needed to bump a few things:

- `simple-configuration-ssm` to >`1.5.5`. This requires aws-sdk v2, which in turn requires us to add credentials objects for both versions in code. This repetition is contained within the AppLoader classes in each project, so it's not too pernicious.
- LanguageTool to `5.3`. This allows us to remove the reference to bintray-hosted `xgboost-predictor`.

## How to test

This should be a no-op – the project should build and work as before.

Tested in CODE.


tjsilver (approved):
👍 builds on TC!

PR #141, opened 2021-04-29
Revert "Remove bintray resolver"

jonathonherbert:
Reverts guardian/typerighter#140 – I merged before considering the impact of the LanguageTool bump. It should be fine but we should check for any changes to the native LT rules before merging.


tjsilver (approved):
Good point!

PR #142, opened 2021-04-29
Add rule checker script

jonathonherbert:
## What does this change?

LanguageTool has default rules that we use, and as we upgrade LT, these could change underneath us.

This PR adds a script to help us see what's changed. Given two grammar.xml files from different LanguageTool versions (an [example](https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml)), and a comma-separated list of rules, it should identify what's been changed/removed.

 I'll leave the rest of the description to the script, in the hope that it's self-describing, and gather feedback!

## How to test

Run the script against two files with a list of rules – for example, [this file](https://github.com/languagetool-org/languagetool/blob/v5.3/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml) (5.3), [this file](https://github.com/languagetool-org/languagetool/blob/v4.3/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml) (4.3) and this list of rules (our current set): 

`NO_SPACE_CLOSING_QUOTE,PUBIC_X,IN_PRINCIPAL,CURRENCY,CURRENCY_SPACE`

You should see output finding that one rule has changed.


SHession (approved):
Tested this locally, code looks good and works as described. Is it worth mentioning this script in the README or signposting it's usage somewhere? How do you envision this script being used, and how would people know to use it?

EDIT: It might also be more congruous to read the rule_ids from a file.

jonathonherbert:
Agreed on all points, great feedback.

PR #143, opened 2021-04-29
Revert "Revert "Remove bintray resolver""

jonathonherbert:
## What does this change?

Un-reverts #141 (implements #140) – we've checked the LanguageTool rules with a script (#142) and all appears OK.

See original #140 for description and testing.


twrichards (approved):
i see you've tested so 👍

PR #148, opened 2021-05-13
Upgrade to Bionic AMI

SHession:
## What does this change?

This PR changes the riff-raff config to use the newly created bionic-arm recipe. This was necessary as the current Xenial recipe is failing to bring up new boxes, and the OS has now left LTS.

## How to test

Deploy this to CODE, check the boxes come up successfully and Typerighter works as expected. 

This should be a no-op from a user perspective.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -14,7 +14,7 @@ deployments:
     app: typerighter
     parameters:
       amiTags:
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
I can see a happy box in CODE 👍

PR #164, opened 2021-09-17
Move gu-cdk for rule management app to Patterns

jonathonherbert:
## What does this change?

Moves the gu-cdk for the rule management app to the new `Patterns` classes. This significantly reduces the boilerplate required.

## How to test

This should be a no-op – the service should continue to run as normal.

To test, you could try running `yarn synth` in the `./cdk` folder, and cloudforming the resulting stack in `./cdk.out`. The result should be a working service.

Tested in CODE.

Reduced LoC and more comprehensible cdk.

### DIFF
--- a/cdk/.gitignore
+++ b/cdk/.gitignore
@@ -1,5 +1,6 @@
 *.js
 !jest.config.js
+!jest.setup.js
 *.d.ts
 node_modules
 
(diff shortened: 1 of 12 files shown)


jonathonherbert:
Closing in favour of #168

PR #168, opened 2021-11-23
Use CDK, not cloudformation, for all apps

jonathonherbert:
## What does this change?

Moves all our apps to the latest version of CDK. This involves a few high-level changes:

- Migrating the checker app to CDK
- Using the latest Patterns where they're available
- Using CDK for our DNS
- Making the appropriate changes to our deployment

We think it'll be easier to spin up a new stack in parallel to the old, and then point consumer apps to the new subdomains. We've taken this opportunity to rationalise the subdomains:

`(api. -> checker.)typerighter`
`({no subdomain} -> manager.)typerighter`

## How to test

This PR is ... larger than I'd like 😅 There wasn't an easy way to make this change incrementally, so I've gone for a single PR. I'd be super happy to pair on a review.

Deploy the app to CODE. You should see services on the following subdomains behaving as expected:

`checker.typerighter.code.dev-gutools.co.uk`
`manager.typerighter.code.dev-gutools.co.uk`

To exercise the service, change the configurating in a consuming service, e.g. Composer CODE, to point to the new URL. You should see it working correctly.

Tested in Composer CODE. This is currently deployed, so you should be able to test as-is.

- The app continues to work as expected
- CDK! Fewer LOC, clearer configuration.

There shouldn't be any denial of service, as we're pointing consuming services to a new URL.


cdk/package.json
```diff
@@ -9,23 +9,28 @@
     "watch": "tsc -w",
     "test": "jest",
     "cdk": "cdk",
-    "synth": "cdk synth rule-manager > ./cfn/rule-manager.cfn.yaml && cdk synth rule-manager-db > ./cfn/rule-manager-db.cfn.yaml"
+    "synth": "cdk synth"
```
akash1810, line 12: Would suggest [adding](https://github.com/guardian/cdk-cli/blob/dd3460d17937a9de3c24bfe11f8bad0bba0f66e5/src/template/package.json#L16) `--path-metadata false --version-reporting false` flags to the `synth` command to reduce bloat in the resulting CFN template.

cdk/lib/index.ts
```diff
@@ -0,0 +1,302 @@
...
+    const pandaAuthPolicy = new GuGetS3ObjectsPolicy(this, "PandaAuthPolicy", {
+      bucketName: "pan-domain-auth-settings",
+    });
+
+    const checkerDomainCODE = "checker.typerighter.code.dev-gutools.co.uk";
+    const checkerDomainPROD = "checker.typerighter.gutools.co.uk";
+    const managerDomainCODE = "manager.typerighter.code.dev-gutools.co.uk";
+    const managerDomainPROD = "manager.typerighter.gutools.co.uk";
+
+    const stageLookup = new CfnMapping(this, "LowercaseStageLookup", {
```
akash1810, line 78: Could we use [`withStageDependentValue`](https://github.com/guardian/cdk/blob/aadfe79af61bed401393e658facdb5e1eba22706/src/constructs/core/stack.ts#L77-L87) from `GuStack` here?

cdk/lib/index.ts
```diff
@@ -0,0 +1,302 @@
...
+          PROD: managerDomainPROD,
+          CODE: managerDomainCODE,
+        },
+      },
+    });
+
+    const typerighterBucketName = `typerighter-${stageLookup.findInMap(
+      "lowercase",
+      this.stage
+    )}`;
```
akash1810, line 98: I'm increasingly of the opinion that in this scenario we should have a single bucket that is sharded by stage, rather than a bucket per stage (akin to an account's artifact bucket).

Would love to get your opinion on that/hear why you have a bucket per stage.

akash1810: cc @jfsoul @jacobwinch @nicl as we were discussing this recently.

cdk/lib/index.ts
```diff
@@ -0,0 +1,302 @@
...
+      },
+      monitoringConfiguration: {
+        noMonitoring: true,
+      },
+      roleConfiguration: {
+        additionalPolicies: [pandaAuthPolicy],
+      },
+    });
+
+    const ruleManagerDnsRecord = new GuDnsRecordSet(this, "manager-dns-records", {
```
akash1810, line 127: YES!

akash1810 (commented):
This PR is delightful!

cdk/lib/index.ts
jonathonherbert: Makes sense, and an easy change to make.

cdk/lib/index.ts
jonathonherbert: We can! And we will! Thanks for making me aware of this very useful API.

jonathonherbert:
Thanks for the review, @akash1810! I hope to have addressed those comments above.

cdk/lib/index.ts
```diff
@@ -0,0 +1,307 @@
...
+          pandaAuthPolicy,
+          new GuPutCloudwatchMetricsPolicy(this),
+        ],
+      },
+    });
+
+
+    const typerighterBucket = new GuS3Bucket(this, "typerighter-bucket", {
+      bucketName: typerighterBucketName,
+    });
```
akash1810, line 185: This won't work quite the way we'd expect! As we're explicitly setting the bucket name, we're creating a scenario where CODE and PROD will _both_ attempt to create a bucket of this name. This isn't legal as the bucket name is a unique identifier, thus can only be created once.

Alternatives would be to create a bucket per stack (like we were previously doing in this branch) or create the bucket outside of this stack.

Apologies, as this is the opposite of what was previously recommended. A Discussion has been opened [here](https://github.com/guardian/recommendations/discussions/74).

jonathonherbert: Ah, yep! Should've spotted – what's best? Can't help but feel that two buckets are preferable, although not ideal, as they keep everything as IaC, but keen to hear your take.

jonathonherbert: Amended to `typerighter-${stage}` per the way that discussion is going – I do prefer keeping bucket creation in VCS.

akash1810 (approved):

PR #170, opened 2021-12-17
Bump Play to 2.8.11, SBT to 1.5.7

jonathonherbert:
## What does this change?

Bumps Play to 2.8.11, SBT to 1.5.7.

## How to test

Run locally. (Do not deploy to CODE – it's waiting for #168!) The application should work as before.


script/set-node-version
```diff
@@ -0,0 +1,12 @@
+nvm_available() {
```
SHession, line 1: It looks like this scripts is missing a `Shebang`. I think it also has incorrect permissions.

apps/checker/app/matchers/LanguageToolMatcher.scala
```diff
@@ -138,30 +137,30 @@ class LanguageToolFactory(
   }
 
   private def annotateExceptionWithRuleId(e: Throwable, ruleId: String, additionalMessage: String = "") =
-    new Exception(s"Error applying LanguageTool rule `${ruleId}`: ${e.getMessage} ${if (additionalMessage.size > 0) additionalMessage else ""}".trim, e)
+    new Exception(s"Error applying LanguageTool rule `${ruleId}`: ${e.getMessage} ${if (additionalMessage.nonEmpty) additionalMessage else ""}".trim, e)
 }
 
 object LanguageToolMatcher extends MatcherCompanion {
-  def getType = "languageTool"
+  def getType() = "languageTool"
```
SHession, line 144: It feels like we may be a little inconsistent with brackets. [This post ](https://stackoverflow.com/questions/7600910/difference-between-function-with-parentheses-and-without)describes the different between empty parameters and parameterless. It seems like this would be an example of parameterless. Not very familiar with this distinction is apologies in advance if I've misunderstood.

SHession (commented):
Changes look good and I've tested locally to see that it still works as expected. I'm seeing a few depreciation warnings when I compile which we may want to address. I have also left a couple of comments for your consideration.

script/set-node-version
jonathonherbert: It's missing a shebang b/c it's not designed to be directly executed as a script.

I don't think we need executable permissions here, because the file itself is sourced by the shell via another script, not executed.

jonathonherbert: I think there's definitely an argument to add .sh as a suffix to this (and other scripts in this folder) to make it clearer what the files contain, mind!

SHession: Possibly we could use `#!/bin/false`?

jonathonherbert: Ooh, just googled this – perfect! For others who are unsure, according to [the internet](https://stackoverflow.com/a/41849445/3631858), #!/bin/false is a safety mechanism of preventing execution of include-only files.

apps/checker/app/matchers/LanguageToolMatcher.scala
jonathonherbert: I probably did this to satisfy IntelliJ 😅 Happy to remove.

jonathonherbert: Ah, yep, it was asking us to be consistent with a change that's been incorrect for a while – paren methods in the parent trait. Modified upstream.

jonathonherbert:
Thanks – hopefully addressed those comments.

SHession (commented):
Having a few compile issue locally and on TC by the looks of it. Might be off the back of my recommendations 😬

apps/checker/app/matchers/LanguageToolMatcher.scala
jonathonherbert: I'm leaving these changes for now, as they require lots of changes downstream from the fn definitions. Probably better applied globally to project via scalafmt or similar.

jonathonherbert:
Just unearthed a fair few deprecation warnings via `clean` – working through them now.

tjsilver (approved):
Tested in Composer with Typerighter running locally (changed config to point to `.local`) and all worked as expected. A few Scala 3 warnings when compiling but I guess these are for future-proofing as this project is using Scala 2.13..? Great work!
image

tjsilver:
Can confirm that the Scala 3 warnings are no longer since the last push.

PR #197, opened 2022-08-09
Clean up

jonathonherbert:
## What does this change?

Originally intended to be included in #20 – this PR cleans up a few things, reducing some repetition, and adding a few comments.

## How to test

This is a no-op – the tests should pass. Do the changes make things easier to understand?


rhystmills (approved):
Changes looks sensible and app running and testing as expected locally 👍

PR #199, opened 2022-09-12
Remove duplicate lock file from project

jfsoul:
## What does this change?
It appears that the lock file for the `rule-manager-client` project now live in `apps/rule-manager/rule-manager-client`. This would indicate we have a duplicate that can be deleted from its original directory.

This is causing problems for Snyk monitoring of the project because it doesn't understand an orphaned lock file.


jfsoul:
Appears to be get stuck during CI due to inability to publish artifacts. My assumption is that isn't caused by these changes.

jonathonherbert:
Thanks for spotting, @jsoul!

PR #200, opened 2022-09-13
Specify major version of RDS only

jonathonherbert:
## What does this change?

Our RDS database is configured to auto update, and has bumped to 13.4. Meanwhile our CDK is hardcoded to 13.3, and AWS sensibly refuses to downgrade when the CFN is deployed.

This PR specifies a major version for the RDS only, letting it drift up minor versions.

## How to test

Deploy to CODE. The deployment should succeed.

### DIFF
--- a/cdk/lib/index.ts
+++ b/cdk/lib/index.ts
@@ -292,7 +292,7 @@ dpkg -i /tmp/package.deb`,
       vpc: ruleManagerApp.vpc,
       vpcSubnets: { subnetType: SubnetType.PRIVATE_WITH_NAT },
       allocatedStorage: 50,
-      allowMajorVersionUpgrade: true,
+      allowMajorVersionUpgrade: false,
       autoMinorVersionUpgrade: true,
       backupRetention: Duration.days(10),
       engine: DatabaseInstanceEngine.postgres({
--- a/cdk/lib/index.ts
+++ b/cdk/lib/index.ts
@@ -296,7 +296,7 @@ dpkg -i /tmp/package.deb`,
       autoMinorVersionUpgrade: true,
       backupRetention: Duration.days(10),
       engine: DatabaseInstanceEngine.postgres({
-        version: PostgresEngineVersion.VER_13_3,
+        version: PostgresEngineVersion.VER_13,
       }),
       instanceType: "t3.micro",
       instanceIdentifier: `typerighter-rule-manager-store-${this.stage}`,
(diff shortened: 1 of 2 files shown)


cdk/lib/index.ts
```diff
@@ -296,7 +296,7 @@ dpkg -i /tmp/package.deb`,
       autoMinorVersionUpgrade: true,
       backupRetention: Duration.days(10),
       engine: DatabaseInstanceEngine.postgres({
-        version: PostgresEngineVersion.VER_13_3,
+        version: PostgresEngineVersion.VER_13,
```
akash1810, line 299: I can't seem to make this comment on the correct line, but I'd strongly recommend setting `allowMajorVersionUpgrade` above to `false` too to avoid surprises (breaking changes etc.)

jonathonherbert: Agreed – worth encoding this as a default in CDK? (Happy to raise a PR!)

akash1810:
This is failing the build in a confusing way. AFAICT the build steps are successful up to the point of uploading artifacts to Riff-Raff. It also looks like the exit code is 0, which is misleading:

image

This project currently uses [`riffRaffNotifyTeamcity`](https://github.com/guardian/typerighter/blob/05e3b077b192b2d9f05f84027095519a7fb30be4/script/citest#L71) to upload to Riff-Raff. Ultimately, this [calls](https://github.com/guardian/sbt-riffraff-artifact/blob/e6f5e62d8f776b1004f72ed1ea415328fa43ed31/src/main/scala/com/gu/riffraff/artifact/RiffRaffArtifact.scala#L118-L130) the [`publishArtifacts` TeamCity service message](https://www.jetbrains.com/help/teamcity/service-messages.html#Publishing+Artifacts+While+Build+is+in+Progress).

I'm not sure we strictly need this for this project? Might [`riffRaffUpload`](https://github.com/guardian/sbt-riffraff-artifact/blob/e6f5e62d8f776b1004f72ed1ea415328fa43ed31/src/main/scala/com/gu/riffraff/artifact/RiffRaffArtifact.scala#L42) be better? It might surface errors in a more understandable way too 🤞🏽 .

jfsoul:
I tracked down one of the errors in the server logs for a build of my branch (safe to assume it'll be the same for this one too.) I'm not sure it's particularly helpful unfortunately?

image

jonathonherbert:
CI issues addressed in #201.

akash1810 (approved):

PR #201, opened 2022-09-14
ci: Switch to a more direct upload to Riff-Raff

akash1810:
## What does this change?

We've, somewhat randomly (see #200, and #199), started seeing builds fail when uploading artifacts.

The [`riffRaffNotifyTeamcity`](https://github.com/guardian/typerighter/blob/05e3b077b192b2d9f05f84027095519a7fb30be4/script/citest#L71) task is a proxy to [TeamCity's `publishArtifacts` service message](https://www.jetbrains.com/help/teamcity/service-messages.html#Publishing+Artifacts+While+Build+is+in+Progress), which allows:

> You can publish the build artifacts while the build is still running, immediately after the artifacts are built.

In this change, we switch to the `riffRaffUpload` task, which uploads artifacts at the _end_ of a build,  rather than as they are created.

The hope is this fixes the build, or provides more information to diagnose the underlying issue.

## How to test

CI should produce a deployable build.

CI is fixed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -135,3 +135,5 @@ riffRaffArtifactResources := Seq(
   baseDirectory.value / "riff-raff.yaml" -> "riff-raff.yaml",
   baseDirectory.value / "cdk/cdk.out/typerighter.template.json" -> "typerighter-cloudformation/typerighter.template.json"
 )
+
+riffRaffManifestProjectName := s"Editorial Tools::Typerighter"
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,5 @@
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.11")
-addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.9")
+addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 addSbtPlugin("io.gatling" % "gatling-sbt" % "3.0.0")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-less" % "1.1.2")
--- a/script/citest
+++ b/script/citest
@@ -68,7 +68,7 @@ buildClients() {
 buildSbt() {
   (
     cd $ROOT_DIR
-    sbt clean compile test riffRaffNotifyTeamcity
+    sbt clean compile test riffRaffUpload
   )
 }
 


project/plugins.sbt
```diff
@@ -1,5 +1,5 @@
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.11")
-addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.9")
+addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
```
akash1810, line 2: Updating to get some nice defaults - https://github.com/guardian/sbt-riffraff-artifact/pull/63.

build.sbt
```diff
@@ -135,3 +135,5 @@ riffRaffArtifactResources := Seq(
   baseDirectory.value / "riff-raff.yaml" -> "riff-raff.yaml",
   baseDirectory.value / "cdk/cdk.out/typerighter.template.json" -> "typerighter-cloudformation/typerighter.template.json"
 )
+
+riffRaffManifestProjectName := s"Editorial Tools::Typerighter"
```
akash1810, line 139: When using `riffRaffNotifyTeamcity`, the project name defaults to the TeamCity project name (Editorial Tools::Typerigher).

When using `riffRaffUpload`, the project name defaults to the SBT project name (typerighter).

We need to override the default to remain consistent.

project/plugins.sbt
jonathonherbert: Presume this does not affect this project? I can't see any values removed in this PR.

build.sbt
jonathonherbert: It's arguably nicer to store this in code, anyhow – it means this project will turn up in a search if people are working backwards from a deploy.

jonathonherbert (approved):
Thanks for this. I'm quite confused that this has begun to fail. I'll take a look at `riffRaffNotifyTeamcity`, as I don't yet understand the difference between this and the command that replaces it, but I can see this PR building, so ✅ . Is this happening in other projects?

akash1810:
> Thanks for this. I'm quite confused that this has begun to fail. I'll take a look at `riffRaffNotifyTeamcity`, as I don't yet understand the difference between this and the command that replaces it, but I can see this PR building, so ✅ . Is this happening in other projects?

It's pretty confusing! @jfsoul has done [some investigating](https://github.com/guardian/typerighter/pull/200#issuecomment-1246507260) however it points to an internal error. Regarding other projects being affected, good question, not sure! We'll need to [check](https://github.com/search?q=org%3Aguardian+riffRaffNotifyTeamcity&type=Code).

PR #204, opened 2022-09-23
Calculate percentage complete and return in check result

rebecca-thompson:
## What does this change?
This updates the chunked/streamed response in the Typerighter API to return an indication of what percentage of jobs have been completed in a check request. This percentage is currently calculated on the client-side and by calculating this value on the server we will be able to simplify our client-side code.

## How to test
Deploy and run this branch locally and run the accompanying branch in [prosemirror-typerighter](https://github.com/guardian/prosemirror-typerighter/pull/233). You should be able to see the loading bar in the right hand side results panel continue to work as expected.

### DIFF
--- a/apps/checker/app/services/MatcherPool.scala
+++ b/apps/checker/app/services/MatcherPool.scala
@@ -1,24 +1,21 @@
 package services
 
-import java.util.concurrent.ConcurrentHashMap
-
-import net.logstash.logback.marker.Markers
-
-import scala.jdk.CollectionConverters._
-import scala.concurrent.{ExecutionContext, Future, Promise}
-import scala.util.{Failure, Success, Try}
-import scala.concurrent.duration._
-import model.{BaseRule, Category, Check, CheckResult, RuleMatch, TextBlock}
-import utils.{Matcher, RuleMatchHelpers}
+import akka.NotUsed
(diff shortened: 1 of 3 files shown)


apps/checker/app/services/MatcherPool.scala
```diff
@@ -123,24 +117,35 @@ class MatcherPool(
...
+
+      logger.info(s"Created $totalJobCount jobs for request with id: ${query.requestId}")
 
       val eventualResponses = jobs.map(offerJobToQueue)
       val responseStream = Source(eventualResponses).mapAsyncUnordered(1)(identity)
 
       responseStream.map {
-        case (job, matches) => CheckResult(job.categoryIds, job.blocks, matches)
+        case (job, matches) =>
+          jobsCompleted.incrementAndGet()
```
jonathonherbert, line 137: Haha, can't believe there's not just an `increment` 🔍 🤔

jonathonherbert (approved):
I'm unable to test as Cisco is playing up, but the code looks good – I'd forgotten that figuring this out is easier on the server, where we know more about how a request is divided into jobs. Feel free to merge, or if you'd prefer another pair of 👀, happy to give this a spin on Monday or defer to another dev.

jonathonherbert:
Tested locally, works as expected.

PR #220, opened 2022-12-09
Clean up unused vars

jonathonherbert:
## What does this change?

A bit of housekeeping: turns on warnings for unused vars, makes those warnings fatal, and cleans up.

## How to test

This should be a no-op: project should build, tests should pass.


jonathonherbert:
Now redundant as we've made many changes.

PR #229, opened 2023-03-01
Move common dependencies to common-lib

aracho1:
_co-authored-by: @rhystmills, @jonathonherbert, @ParisaTork, @phillipbarron_   

## What does this change?

Adds /rules and /refresh endpoints to the rule manager service. They behave in a similar way to the /rules and /refresh endpoints in the checker service – but pass JSON, rather than whole pages, in anticipation of them being called from a single page app.

## How to test

Running locally or in code, hit the /refresh or /rules endpoints with POST and GET requests respectively. /refresh should pull the rules from the Google sheet of the stage you're operating in (the ID of the sheet will be in AWS Parameter Store at `/${STAGE}/flexible/typerighter/sheetId`), write them to S3, and return them all as JSON. /rules should read the existing rules from S3, and return them as before.

(These requests will need to carry pan-domain-auth cookies, so it might be worth copying an existing request to the management service from your browser's network tab, and using 'copy as curl' or similar to easily grab cookie information.)

Screenshot 2023-03-02 at 17 12 07

Developers are able to refresh rules from the Google sheet, and read them from S3.

Additional work to add the UI for non-developer users to come.

Once we ship this, there'll be two ways of writing rules to S3. I don't think that's a problem, but we should deprecate rule management in the checker as soon as we ship something that lets us manage rules from the rule manager.


jonathonherbert:
This deploys fine but refreshing/getting rules doesn't yet work – we need to add permission for the rule-manager to access the rules json.

PR #231, opened 2023-03-07
Create rules table in the postgres database

aracho1:
_co-authored-by: @ParisaTork, @jonathonherbert, @phillipbarron, @rhystmills_   
## What does this change?

An initial pass at a table to contain our rules. It's intended to model the rules pretty match exactly as they're represented in the Google sheet.

We anticipate this model changing as we work on the rule manager, but this gets us off the ground.

This PR does not include code to add user-facing interaction with the DB, but the tests should prove that we're able to connect and read/write models.

## How to test

- Run the tests locally, ensuring that the database is present with `docker compose up`. Do they pass?


jonathonherbert (approved):
Tested in CODE, application deploys successfully and I can see the schema in the CODE DB.

PR #269, opened 2023-03-15
Remove PR frequency from Scala Steward

ParisaTork:
## What does this change?

Remove PR frequency from Scala Steward

## How to test

Should generate Scala Steward PRs

## Accessibility

-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -2,10 +2,6 @@
 # Default config settings are here: https://github.com/scala-steward-org/scala-steward/blob/c1aa49d38b3d5fa6adf63f59de240680d7f9135b/docs/repo-specific-configuration.md
 # Check if department-wide Scala Steward GitHub Action has broken on a team’s repo here: https://github.com/guardian/scala-steward-public-repos/actions/workflows/public-repos-scala-steward.yml
 
-# 10am Monday to Friday PR cron
-pullRequests.frequency = "0 10 * * 1-5"
-
-# Groups Scala Steward PRs together, in line with current DevX guidance (Server-side meetup - 1st February 2023)
 pullRequests.grouping = [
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -2,10 +2,6 @@
 # Default config settings are here: https://github.com/scala-steward-org/scala-steward/blob/c1aa49d38b3d5fa6adf63f59de240680d7f9135b/docs/repo-specific-configuration.md
 # Check if department-wide Scala Steward GitHub Action has broken on a team’s repo here: https://github.com/guardian/scala-steward-public-repos/actions/workflows/public-repos-scala-steward.yml
 
-# 10am Monday to Friday PR cron
-pullRequests.frequency = "0 10 * * 1-5"
```
jonathonherbert, line 6: I've come across some configuration docs that seem to suggest that the default here is `@asap`, which is probably not what we want – suspect we won't benefit from the grouping behaviour.

That config also suggests that we can do something like `pullRequests.frequency = "7 days"
 `, which sounds nice and readable – what do you think?

ParisaTork: You'll find `pullRequests.frequency = "7 days"` already in the groupings

jonathonherbert: That doesn't seem to be applying at the moment – scala-steward is still raising ad-hoc PRs, e.g. https://github.com/guardian/typerighter/pull/271.

I wonder whether that's because the 7 day frequency only applies to certain groups? Perhaps a top-level setting will catch everything, which might be what we want.

ParisaTork: That's a fair point - I'll change that now

jonathonherbert (approved):

PR #239, opened 2023-03-15
Revert "Revert "Move common dependencies to common-lib""

jonathonherbert:
Reverts guardian/typerighter#234 and includes https://github.com/guardian/typerighter/pull/235 and a bump to pan-domain-auth to include https://github.com/guardian/pan-domain-authentication/pull/114 – all of which should fix the conflicts with `google-api-client` that led to the revert.

## How to test

- [x] Run the application in CODE. It should deploy as expected.
- [x] Refresh the rules in the checker. The ingest process should work as expected.
- [x] Run a check against a document. It should produce matches as expected. 
- [x] Import rules via the management system (guidance in #229.) Run a check with these newly imported rules. 

Tested all of the above in CODE.

PR #270, opened 2023-03-16
Add ScalaFMT Config, guide & apply to projects. 

phillipbarron:
## What does this change?

Adds config for ScalaFMT and notes to the readme on its use via command line and automatically via the IDE. 
The PR also formats all three Scala projects

## How to test

There have been some linting changes applied as a result of the format. Removal of redundant enclosing blocks in case statements & formatted strings for example & using `nonEmpty` on lists rather than checking for `.size > 0`. 

All tests are passing however, spinning up the app and  verifying it is behaving as expected would be sensible.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,3 @@
+version=3.7.2
+runner.dialect = scala213
+maxColumn = 100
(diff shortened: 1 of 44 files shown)


jonathonherbert (approved):
This is great!

What do we think about running this in CI, and a pre-push githook (perhaps we could install it in the setup script?) – without enforcing these changes we'll drift away from scalafmt perfection.

Finally, shall we wait until #231 and #239 are merged before merging this? Suspect the conflicts will ... numerous.

phillipbarron:
> 

I added some notes to the card - agree 100% with a pre-commit hook. There is an existing card to move from TeamCity to GHA which might add this, though there is nothing to stop us going that part right away

PR #273, opened 2023-03-16
Write Google spreadsheet rules to DB

aracho1:
## Accessibility

-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)


apps/rule-manager/app/service/DbRuleManager.scala
```diff
@@ -0,0 +1,94 @@
...
+          category = Some(category.name),
+          description = Some(description),
+          replacement = None,
+          ignore = false,
+          googleSheetId = Some(id)
+        )
+    }
+  }
+
+  def dbRuleToBaseRule(rule: DbRule): BaseRule = {
```
phillipbarron, line 42: There's a lot of `get` in here - we cannot construct a rule where one of these values is missing so perhaps this is fine but might we want to wrIte what we can but trigger some alert where we have not been able to write some rule do to the absence of a value on which we rely?

Maybe one to look at later, perhaps adding a todo will suffice for now?

Partial success would need to modelled in the response too, so perhaps all or nothing is the simpler option

phillipbarron (approved):
Looks good - added a comment in the rule manager - we talked about refactoring all the `get` out but this need not be a blocker

apps/rule-manager/app/service/DbRuleManager.scala
davidfurey: I can understand that `GoogleSheetId` is a temporary concept, but if `category` and `description` are actually required, why are they optional in the type system?

davidfurey: If you change the signature of the method you can avoid the `gets`:
```
  def dbRuleToBaseRule(rule: DbRule): Option[BaseRule] = PartialFunction.condOpt(rule) {
    case DbRule(_, "regex", pattern, replacement, Some(category), _, Some(description), _, _, Some(googleSheetId), _, _) =>
      RegexRule(
        id = googleSheetId,
        category = Category(id = category, name = category),
        description = description,
        suggestions = List.empty,
        replacement = replacement.map(TextSuggestion(_)),
        regex = new ComparableRegex(pattern)
      )
    case DbRule(_, "languageTool", pattern, _, Some(category), _, Some(description), _, _, Some(googleSheetId), _, _) =>
      LTRuleXML(
        id = googleSheetId,
        category = Category(id = category, name = category),
        description = description,
        xml = pattern
      )
  }
```

davidfurey: "we should probably be able to save a rule without a category or a description" - makes sense, I hadn't thought about work in progress.

PR #278, opened 2023-03-21
update rule db instance type to db.t4g.micro

phillipbarron:
## What does this change?

Changes the Rule DB instance type to `db.t4g.micro`

## How to test

Deploy to code and check that the db is up and running.


phillipbarron:
Tested on code - looks fine - As we're not actually using this yet seems safe to push on

PR #279, opened 2023-03-21
Use Vite

jonathonherbert:
## What does this change?

Use Vite, not webpack, for our client side builds. 

## Why?

- Less configuration
- Faster build times via [`swc`](https://swc.rs/)
- A significantly smaller build (>4MB, per observations in #277, to ~500kb pre-gzip)
- HMR during development:

edit-realtime

Previously, Scala was recompiling as a result of asset changes, so this is a significant improvement to the feedback loop when developing server side.

## How to test

Run locally by running `./scripts/start-manager`. The app should run locally as expected, backgrounding the build server. Try changing something in the components. The app should update in realtime.

Test the build process – which is quite different from the local dev process! – by deploying to CODE. The app should run as expected.

- [x] Tested in CODE.

A better feedback loop for devs, and a smaller bundle size in production.


phillipbarron (approved):
Love it. Tested locally, all working 🎉

PR #282, opened 2023-03-22
Use Localstack s3 bucket when running locally

rhystmills:
Co-authored with @aracho1 

## What does this change?

This PR adds a localstack S3 bucket which will be written to and read from when running checker or rule manager locally (rather than writing to the CODE S3 bucket).

The [`localstack/init-aws.sh`](https://github.com/guardian/typerighter/compare/main...ac-rm/localstack-bucket#diff-75f2bd3e10422803c8e7553aad876b0a373155e6b7362a10e8839e975b503913) script is copied to the Docker localstack instance due to the following rules in `docker-compose.yml`: 
``` - "./localstack/init-aws.sh:/etc/localstack/init/ready.d/init-aws.sh"```

Docker will run the script on initialisation and create the local bucket.

`pan-domain-authentication` requires its own s3 client because we don't run it locally.

## How to test

1. Run the app locally - `./script/start`. Ensure docker is running.
2. Visit https://checker.typerighter.local.dev-gutools.co.uk/ and head to the 'Rules' tab.
3. Click 'Refresh rules'
4. Are rules pulled into your checker app? Do they include the nonsense rule 'advantagemently'?
5. Consider checking the localstack log output in Docker to see if it's handling s3 requests

image

### CODE REVIEW

File: apps/rule-manager/app/AppComponents.scala [hunk 1]
```diff
@@ -35,8 +37,23 @@ class AppComponents(
...
+       .standard()
+      .withCredentials(creds)
+      .withEndpointConfiguration(new EndpointConfiguration("http://localhost:4566", Regions.EU_WEST_1.getName))
+      // This is needed for localstack
+      .enablePathStyleAccess()
+      .build()
+  }
 
-  private val s3Client = AmazonS3ClientBuilder
+  private val pandaS3Client = AmazonS3ClientBuilder
```
jonathonherbert, line 56: As above, if appropriate.

jonathonherbert (approved):
This looks great! A few non-blocking comments.

✅ I can see the docker image come up with the correct bucket
✅ I can see rules written to that bucket in the docker logs
✅ I can see `advantagemently` in local checker

One thing's a bit odd – the same rules (DEV) seem to be picked up by CODE, in that I can see advantagemently there too. I wonder if our CODE box has odd ideas about its stage? I'll take a closer look tomorrow.

apps/rule-manager/app/AppComponents.scala
```diff
@@ -57,20 +74,20 @@ class AppComponents(
...
     bucketName = "pan-domain-auth-settings",
     settingsFileKey = s"$stageDomain.settings",
-    s3Client = s3Client
+    s3Client = pandaS3Client
   )
 
   val stage = identity match {
     case identity: AwsIdentity => identity.stage.toLowerCase
-    case _                     => "code"
+    case _: DevIdentity => "local"
```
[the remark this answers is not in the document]
rhystmills: (Sealed trait/enum definitely sounds sensible)

apps/rule-manager/app/AppComponents.scala
rhystmills: (Also done)

jonathonherbert (approved):

PR #283, opened 2023-03-22
Support CI via Github Actions

aracho1:
## What does this change?

This PR build and deploys to RiffRaff via Github Actions.

Following the merge of this PR, we will want to:
1. include `scalaFmt` checks. We should add it as part of the Github action.
2. disable Teamcity's builds to stop duplicates occurring.

## How to test

We're not sure how to test this change until this PR has been merged. However, this branch has been deployed via an action, you can see it in [RiffRaff](https://riffraff.gutools.co.uk/deployment/request) by looking at the builds for `Editorial Tools::Typerighter` (it's build 1297).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,9 +22,3 @@ jobs:
         with:
           java-version: "8"
           distribution: "corretto"
-      - name: test and check formatting
-        run: |
-          sbt -v test scalafmtCheckAll
-      - name: build uberjar
-        run: |
-          sbt -v assembly
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,3 +22,8 @@ jobs:
         with:
           java-version: "8"
           distribution: "corretto"
+      - name: Build the stack
+        run: docker-compose up
+      - name: test and check formatting
+        run: |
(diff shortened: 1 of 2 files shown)


Fweddi (approved):
LGTM - we might want to also allow CI to run on pushes to non-main branches (though I know in the past this has been mentioned as cost ineffective)

twrichards:
> LGTM - we might want to also allow CI to run on pushes to non-main branches (though I know in the past this has been mentioned as cost ineffective)

Free for public repos (unless using special instance/runner types), better to have builds available for your branches without needing a pull request (worth noting too that pull request triggered actions also use the special merge branch which isn't strictly the branch in question but instead a simulation of the branch merged with the base branch [usually main] which is not very transparent, so I vote for for build branches and not pull requests.

twrichards:
@rhystmills you were asking about how to run on all branches, here's an example... https://github.com/guardian/pinboard/blob/main/.github/workflows/ci.yml#L4-L8

rhystmills:
> @rhystmills you were asking about how to run on all branches, here's an example... https://github.com/guardian/pinboard/blob/main/.github/workflows/ci.yml#L4-L8

Thanks Tom. I agree push works better than PR - that's updated too.

PR #289, opened 2023-03-28
On dump, ensure rules are removed before new rules are added

jonathonherbert:
## What does this change?

On dump, ensure rules are removed before new rules are added.

## How to test

- The automated test should pass.
- CODE rule ingestion should now work as expected.


phillipbarron (approved):
❤️

PR #293, opened 2023-03-29
Update package lock. Possible merge issue resulting in missing references @emotion/hash@^0.9.0 which caused snyk failure

phillipbarron:
The Snyk workflow runs against main so a PR to main can fail post merge - might be sensible to change this bahaviour so we can tell ahead of merge if there is an issue. 
This PR updates the Package-lock in the client package to resolve the Snyk issue

### DIFF
--- a/apps/rule-manager/client/package-lock.json
+++ b/apps/rule-manager/client/package-lock.json
@@ -125,9 +125,9 @@
       }
     },
     "node_modules/@elastic/eui": {
-      "version": "76.2.0",
-      "resolved": "https://registry.npmjs.org/@elastic/eui/-/eui-76.2.0.tgz",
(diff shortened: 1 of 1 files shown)


phillipbarron:
> Could I suggest using [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) over [`npm i`](https://github.com/guardian/typerighter/blob/711f737871ebc381a9efc842c494e3bae7b0892a/.github/workflows/ci.yaml#L34) to install dependencies during CI.
> 
> With `npm ci`, the build would fail if the `package-lock.json` is out of date/sync:
> 
> > * If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.
> 
> That is, using `npm ci` would eliminate the need for similar PRs in future.
> 
> `yarn` has a similar feature via the [`--frozen-lockfile` flag](https://classic.yarnpkg.com/lang/en/docs/cli/install/).

Great idea, yep, cheers @akash1810

jonathonherbert (approved):

PR #294, opened 2023-03-29
Remove refresh logic from the Checker app

phillipbarron:
plus a small update to start script in the rules manager app to automatically terminate the app on `ctrl+c`

- Removes Route
- Removes controller logic
- Removes UI elements of refresh from checker.

## What does this change?

After this change, users will no longer be able to trigger a refresh of the rules from the checker app. Instead, the rule manager UI should be used to pull in changes from the google sheet

## How to test

Test that rules are still displayed in the app

We should make sure that all rule maintainers know about the new UI available in the [rule manager app](https://manager.typerighter.gutools.co.uk/)

## Images

Screenshot 2023-03-29 at 12 05 40

Tested locally

### DIFF
--- a/apps/checker/app/AppComponents.scala
+++ b/apps/checker/app/AppComponents.scala
@@ -101,6 +101,7 @@ class AppComponents(
     case identity: AwsIdentity => identity.stage.toLowerCase
     case _: DevIdentity        => "local"
   }
+
   val typerighterBucket = s"typerighter-app-${stage}"
 
   val cloudWatchClient = identity match {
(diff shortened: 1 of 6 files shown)


jonathonherbert (approved):
Tested on CODE, LGTM!

Worth avoiding merging until we've sent comms to CP and we've verified they're happy.

One suggestion: I wonder whether a convenient link from the Checker UI to the management UI would be useful, as I suspect otherwise it'll be a trip back to the inbox to search for the URL, at least the first few times!

PR #296, opened 2023-03-29
Add languagetool default rules – take 2

jonathonherbert:
## What does this change?

Identical to #288, with a merge target of main. The previous PR was mistakenly merged to the branch @ #292. If the repository does not automatically delete branches on merge, the base branch will not automatically change.

## How to test

See #288.


phillipbarron (approved):
See previous approval 😄

PR #299, opened 2023-03-30
Minor styling update to prove a few theme things

jonathonherbert:
## What does this change?

Makes a few more themeing decisions to bring us closer to a Composer-ish style.

These are mainly to prove that we can do a few things –

- theme both sass/emotion-based components with EUi variables, as EUI currently uses a mix (we're stuck overriding in both sass and Emotion contexts for now, but EUI does have a clear roadmap that's seen [significant progress](https://elastic.github.io/eui/#/package/changelog) in the last few months – search that page for `css-in-js conversions`)
- replace icons

It also fixes an issue we were having where Emotion-based globals were overriding sass globals.

|Before|After|
|--|--|
|Screenshot 2023-03-30 at 14 34 23|Screenshot 2023-03-30 at 14 34 48|

## How to test

Take a look. A bit closer to the angular, austere beauty of our other tooling?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,14 +20,14 @@ jobs:
           java-version: "8"
           distribution: "corretto"
       - name: Build the stack
-        run: | 
+        run: |
           docker-compose up -d
       - name: Run tests
         run: |
           sbt -v test scalafmtCheckAll
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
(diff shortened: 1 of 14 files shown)


phillipbarron (approved):
Looks great - that CI saved us from merging without synchronising the lock file 🎉

jonathonherbert:
@phillipbarron yes! Although, I actually can't get package-lock to change when I run npm locally? 🤔

PR #300, opened 2023-04-03
Add additional properties to the rule model

aracho1:
co-authored-by: @jonathonherbert 

## What does this change?
Further to #273, this PR adds the missing properties (_e.g. notes, googleSheetId, forceRedRule, advisoryRule_) to the rule model. SheetsRuleManager now creates dbRule instead of RuleResource. 

When translating rules in the database into the rule resource for the checker, the rule manager filters out any rules that should be ignored (`ignore = true`).

## How to test
- Using the steps mentioned in https://github.com/guardian/typerighter/pull/229, make a POST request to the /refresh endpoint and inspect the database. The rules should contain the new fields and the corresponding data from the Google Sheet.
image

We can refresh rules from the Google sheet with those additional properties, and read them from S3.
Additional work to add the UI is to come.

There is a new test to check that rules are filtered correctly depending on their 'ignore' flag.

### DIFF
--- a/apps/rule-manager/app/service/SheetsRuleManager.scala
+++ b/apps/rule-manager/app/service/SheetsRuleManager.scala
@@ -103,7 +103,7 @@ class SheetsRuleManager(credentialsJson: String, spreadsheetId: String) extends 
         case (Some(id), _, _) if id.isEmpty =>
           Failure(new Exception(s"empty id for rule (row: ${rowNumber})"))
         case (Some(id), _, None) =>
-        Failure(new Exception(s"Rule type ${ruleType} for rule with id ${id} not supported"))
+          Failure(new Exception(s"Rule type ${ruleType} for rule with id ${id} not supported"))
         case (Some(_), None, _) =>
           Failure(new Exception(s"no Ignore column for rule (row: ${rowNumber})"))
         case (Some(id), Some(ignore), Some(ruleType)) =>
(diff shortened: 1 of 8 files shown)


apps/rule-manager/app/service/SheetsRuleManager.scala
```diff
@@ -87,57 +79,54 @@ class SheetsRuleManager(credentialsJson: String, spreadsheetId: String) extends
...
           Success(
             Some(
-              getLTRuleXML(
-                id,
-                row(PatternRuleCols.Pattern).asInstanceOf[String],
-                row(PatternRuleCols.Category).asInstanceOf[String],
-                row(PatternRuleCols.Description).asInstanceOf[String]
+              DbRule.withUser(
+                id = None,
+                ruleType = ruleType.asInstanceOf[String],
```
jonathonherbert, line 111: There's one problem I've found here – because the `ruleType` in our sheet doesn't perfectly match the `ruleType` [defined in our Db](https://github.com/guardian/typerighter/blob/83cfa3e362e609201bb8bd37da20b72877421fdb/apps/rule-manager/app/service/DbRuleManager.scala#L19), rule ingestion fails as soon as we encounter a difference (`regex` is the same, but `lt` -> `languageToolXML` and `lt_core` -> `languageToolCore` differ.)

To fix, let's map between the sheet and DB values here. (We could change them to match, but I think the DB values are clearer!)

If it's not clear how do that, v. happy to pair!

jonathonherbert (commented):
Looks great! Just one change to mention to fix rule ingestion.

apps/rule-manager/app/service/SheetsRuleManager.scala
jonathonherbert: Resolved in [f697f06](https://github.com/guardian/typerighter/pull/300/commits/f697f06c9bb5feebd299ea8e3aba6e3cff5b66b0)

PR #304, opened 2023-04-17
Update architecture diagram

jonathonherbert:
## What does this change?

Updates the architecture diagram to reflect the fact that the management system, not the checker, now writes the rules artefact.

## How to test

Take a look. All make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,18 @@ flowchart LR
   sheet[Google Sheet]
   client[Typerighter client]
   s3[(typerighter-rules.json)]
+  db[(Postgres DB)]
   owner{{Rule owner role}}
   user{{Rule user role}}
 
   sheet--"Get rules"-->manager
+  manager--"Write rules"-->db
+  manager<--"Read rules"--db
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,19 +32,20 @@ Matches contain the range that match applies to, a description of why the match
...
 
-  sheet--"Get rules"-->checker
-  checker--"Write rule artefact"-->s3
+  sheet--"Get rules"-->manager
+  manager--"Write rule artefact"-->s3
   s3--"Read rule artefact"-->checker
   client--"Request matches"-->checker
 
-  owner-."Force checker to re-fetch sheet".->checker
+  owner-."Force checker to re-fetch sheet".->manager
```
Fweddi, line 48: I think this is now `Force manager`, not `Force checker`.

Fweddi (approved):
LGTM, one nit.

We're close to the stage where a reference to the DB would be useful, but that can come in another PR.

jonathonherbert:
Oh, good point! I've added the DB, too 👍

PR #307, opened 2023-04-18
Add creation and update meta to rule model

jonathonherbert:
**NB:** based on https://github.com/guardian/typerighter/pull/306 – please review that first!

## What does this change?

Adds `createdAt`, `createdBy`, `updatedAt` and `updatedBy` to the rule model. They're updated automatically as rules are created and modified.

### This happens in the application logic! Why not write a database trigger to update `updatedAt`?

We could! But we'd still have to write the user data – it made some sense to me to keep all of the logic for our meta in one domain. If there's a preference for a DB trigger, we could do that instead. One advantage would be that there's no possibility of us missing an update as a result of an error in application code.

## How to test

- The automated tests should pass, and you should be convinced they cover both creation and updating cases.
- Try creating and updating rules in Postman, following the instructions in #305. When you create or update rules, are creation/update values correct? E.g. with Postman:

Screenshot 2023-04-18 at 18 12 36

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -11,6 +11,7 @@ import play.api.mvc._
 import service.DbRuleManager
 
 import scala.concurrent.ExecutionContext
+import scala.util.{Success, Failure}
 
 /** The controller that handles the management of matcher rules.
   */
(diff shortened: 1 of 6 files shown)


jonathonherbert:
Ace, ta! If I understand your comment right,  FYI`assertDatesAreWithinRangeMs` is not native to scalatest, there's a `def assertDatesAreWithinRangeMs` in there somewhere 😀

Fweddi:
> Ace, ta! If I understand your comment right, FYI`assertDatesAreWithinRangeMs` is not native to scalatest, there's a `def assertDatesAreWithinRangeMs` in there somewhere 😀

Ah, I meant the pattern / idea, not the method itself!

jonathonherbert:
Gotcha, yeah sometimes useful to move assertions into their own little helpers 🧹 ✨

PR #308, opened 2023-04-19
Remove warnings, and turn on `-Xfatal-warnings`

jonathonherbert:
## What does this change?

Quick tidy up to remove a few warnings. Turns on `-Xfatal-warnings` to ensure we don't introduce new ones.

## How to test

Compilation should succeed, which with the given config confirms there are no longer any warnings.


davidfurey:
I've started reviewing this but the sand seems to be shifting under my feet. Should it be a draft?

jonathonherbert:
Thanks for looking, @davidfurey. It's ready now, but it's probably sensible to move to draft and get through our backlog of recent work first – suspect it might cause some fun rebases if merged in front of that work.

apps/checker/app/model/Check.scala
```diff
@@ -46,5 +46,5 @@ case class CheckResult(
     categoryIds: Set[String],
     blocks: List[TextBlock],
     matches: List[RuleMatch],
-    percentageRequestComplete: Option[Float] = None
```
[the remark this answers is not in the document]
jonathonherbert: We're already rounding to an `Int` upstream, and converting from Int back to Float for the sake of this interface gave a precision warning. Better to just avoid the conversion.

jonathonherbert:
That backlog's now on `main`, and this is ready for review.

apps/checker/app/services/MatcherPool.scala
```diff
@@ -210,13 +210,14 @@ class MatcherPool(
...
 
-    queue.offer(job).collect {
+    queue.offer(job) match {
       case Dropped =>
         failJobWith(job, "Job was dropped from the queue, as the queue is full")
       case QueueClosed =>
         failJobWith(job, s"Job failed because the queue is closed")
       case QueueFailure(err) =>
         failJobWith(job, s"Job failed, reason: ${err.getMessage}")
+      case _ => // Do nothing if job is accepted
```
davidfurey, line 220: Outside the scope of this PR, but would it be useful to log that the job has been accepted?

davidfurey (approved):
🥳

apps/checker/app/services/MatcherPool.scala
jonathonherbert: 3782a14f1248608f0da8fed17ea4c84f9a78a31b adds this – it's only little, and it's good to be explicit for every case (what if we were dropping log lines?)

PR #310, opened 2023-04-20
Add revisionId to rule model

jonathonherbert:
## What does this change?

Adds a revisionId to the rule model. It's incremented in the application logic when we call `save`.

## How to test

- The automated tests should pass – they check for the increment.
- Create and update a rule via the instructions in #305. You should see a revisionId in the output, which should increment on every update.


Fweddi (approved):
All LGTM. Zero-indexing the revision ID did throw me off at first, but semantically it makes sense (even if in my head revision and version are synonymous).

PR #311, opened 2023-04-21
ADR for rule audit implementation

jonathonherbert:
## What does this change?

Adds an ADR for the rule audit implementation.

## How to test

Have a read. Comments/opinions will be useful in informing a decision!


docs/decision-records/003-audit-rules.md
```diff
@@ -0,0 +1,44 @@
...
+There are a few ways we might model the data necessary to support these features in the database.
+
+# Positions
+
+### 1. Create a separate table, `rules_history`, populated with older versions of rules as rules are edited.
+
+- (+) Reading and writing rules does not change – the `rules` table is canonical
+- (-) We must maintain a duplicate of the rules table. The schema of that table must be kept in sync with the `rules` table.
+- (-) The mechanism for writing rules into `rules_history` introduces complexity:
+  1. Use a trigger. There's a [pSQL implementation of the temporal tables DB extension](https://github.com/nearform/temporal_tables) which looks well-used and tested, but future users may encounter edge-cases which are hard to debug.
```
itsibitzi, line 21: I think the complexity introduced by this automation could dwarf the complexity introduced by manually managing your history. There is some benefit in having your table state visible in the queries.

FWIW the language is `PL/pgSQL` (as in, **P**rogramming **L**anguage, **P**ost**g**res **SQL**, the Postgres specific dialect of SQL that you can use in functions). `psql` is the commandline tool for interacting with an instance.

jonathonherbert, line 21:
```suggestion
  1. Use a trigger. There's a [PL/pgSQL implementation of the temporal tables DB extension](https://github.com/nearform/temporal_tables) which looks well-used and tested, but future users may encounter edge-cases which are hard to debug.
```

docs/decision-records/003-audit-rules.md
```diff
@@ -0,0 +1,44 @@
+# Rule manager: implementing rule edit history
+
+## Context
+
+We'd like to store a history of rule edits. The user stories we'd like to satisfy with this feature are:
+
+- As a rule owner, I'd like to be able to see the previous edits made to a rule, including
+  - what changed, and
```
paperboyo, line 8: And “by whom”?

jonathonherbert:
This no longer applies, as we've made a slightly different architectural choice for rule history.

PR #315, opened 2023-04-26
googleSheetId -> externalId, which is now added to new rules

jonathonherbert:
## What does this change?

Changes the `googleSheetId` to `externalId`, and makes it compulsory, with externalIds generated when new rules are created.

## How to test

- Ingesting rules should preserve their IDs and display them in the grand list of rules.
- Try creating a new rule, following the instructions in #305. Does it have an `externalId`?


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 
 services:
     postgres:
-        image: postgres:10.7-alpine
+        image: postgres:13.7-alpine
```
jonathonherbert, line 5: Not strictly necessary w/ this PR (initially bumped for postgres' UUID() type), but does bring us closer in line w/ production DB.

davidfurey: I suppose it would be nice for this to be kept in sync with https://github.com/guardian/typerighter/blob/ffd616566e16c3e2fc0f00bafc6ceecd35b33563/cdk/lib/index.ts#LL281C40-L281C46

apps/rule-manager/app/db/DbRule.scala
```diff
@@ -140,7 +141,7 @@ object DbRule extends SQLSyntaxSupport[DbRule] {
       description: Option[String] = None,
       ignore: Boolean,
       notes: Option[String] = None,
-      googleSheetId: Option[String] = None,
+      externalId: String = UUID.randomUUID().toString(),
```
davidfurey, line 144: Would it be better to keep this as `Option[String]` and let the database generate the default value?

davidfurey (commented):
I haven't tested the change, but it looks reasonable to me

apps/rule-manager/conf/evolutions/default/5.sql
```diff
@@ -0,0 +1,20 @@
...
+  ALTER COLUMN external_id SET NOT NULL;
+
+CREATE UNIQUE INDEX rules_external_id_index ON rules(external_id);
+
+-- !Downs
+
+DROP INDEX rules_external_id_index;
+
+ALTER TABLE rules
+  ALTER COLUMN external_id drop NOT NULL;
```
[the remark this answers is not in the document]
jonathonherbert: Thanks for spotting! [41882e1](https://github.com/guardian/typerighter/pull/315/commits/41882e1517cb8109bc6b7ee2cd86ee02b5303f6f)

davidfurey (approved):

PR #316, opened 2023-04-27
rename `baseRule` -> `checkerRule` and `ruleResource` -> `checkerRuleResource`

aracho1:
## What does this change?
Renaming
-  `baseRule` -> `checkerRule`
-  `ruleResource` -> `checkerRuleResource`

to make clearer their usage / purpose and distinction with the dbRule.

Since these are simple name changes, there should be no changes to how the rule models / application work.

## How to test
- Everything works same as before.
- All the automated tests should pass.

- More clear distinction between different types of rules and their purposes

Nothing I can think of.


jonathonherbert (approved):
LGTM!

PR #318, opened 2023-05-02
Add create rule form (first pass)

rhystmills:
Co-authored-by @Fweddi 

## What does this change?

This PR adds a 'create rule' form. Currently the form is very permissive, and only the 'rule' is required.

image

The DB model for categories and and tags are subject to change (tags also currently aren't save to the database, PR-pending). In this PR, categories are a hardcoded list, in the future we'd want to use an endpoint, drawing from existing tags. We'd also want tags to be drawn from the database in the future, with users able to add their own tags.

The position of the form in this PR is subject to change - we've tried to cover the core functionality of the form (managing form data and submitting to the backend) as the UI will be finalised later.

## How to test

With Composer credentials, run `typerighter` locally using `./script/start`. Can you add a new rule via the form as intended?

## Accessibility

-   [X] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)

### DIFF
--- a/apps/rule-manager/client/src/ts/components/CategorySelector.tsx
+++ b/apps/rule-manager/client/src/ts/components/CategorySelector.tsx
@@ -3,23 +3,35 @@ import {EuiFormRow, EuiComboBox} from "@elastic/eui";
 import React from "react";
 import { PartiallyUpdateRuleData, RuleFormData } from "./RuleForm";
 
+export type MetadataOption = {label: string}
 export const CategorySelector = ({ruleData, partiallyUpdateRuleData}: {
     ruleData: RuleFormData,
     partiallyUpdateRuleData: PartiallyUpdateRuleData,
 }) => {
-    type Category = {label: string}
-    const categories: Category[] = [
-        {
-            label: 'Category A',
-        },
-        {
-            label: 'Category B',
-        }
+    const categories: MetadataOption[] = [
+        {label: "Check this"},
+        {label: "Guardian convention"},
+        {label: "Style guide and names"},
+        {label: "General"},
+        {label: "Names"},
+        {label: "Typos"},
+        {label: "Tokyo 2020: General"},
+        {label: "Tokyo 2020: Diving"},
+        {label: "Tokyo 2020: Equestrian"},
+        {label: "Tokyo 2020: Cycling"},
+        {label: "Tokyo 2020: Rugby"},
+        {label: "Tokyo 2020: Boxing"},
+        {label: "Tokyo 2020: Football"},
+        {label: "Tokyo 2020: Swimming"},
+        {label: "Coronavirus"},
+        {label: "Typography"},
+        {label: "Dates"},
+        {label: "Style Guide"}
     ]
     const [options, updateOptions] = useState(categories);
     // This is an array in order to match the expected type for EuiComboBox, but 
     // it will never have more than one category selected
-    const [selectedCategory, setSelectedCategory] = useState<Category[]>([]);
+    const [selectedCategory, setSelectedCategory] = useState<MetadataOption[]>([]);
 
     const onChange = (selectedOption) => {
         setSelectedCategory(selectedOption);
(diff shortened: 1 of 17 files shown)


jonathonherbert (commented):
Looks really good! A few comments.

Generally, one other thing I've found – creating a rule without tags gives `Uncaught TypeError: Cannot read properties of undefined (reading 'toString')`.

apps/rule-manager/client/src/ts/components/LineBreak.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export const LineBreak = () => <hr style={{
```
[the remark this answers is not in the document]
rhystmills: I have now done so.

rhystmills:
I believe I've sorted that tag problem, thanks for the review!

apps/rule-manager/client/src/ts/components/CategorySelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+        if (!normalizedSearchValue) {
+            return;
+        }
+
+        const newOption = {
+            label: searchValue,
+        };
+
+        // Create the option if it doesn't exist.
```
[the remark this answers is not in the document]
rhystmills: I've set tags and categories to use a hardcoded list for now. The hardcoded tags are very inconsistent in terms of case etc but are what's in the sheet currently. We can revisit later after a tidy up!

apps/rule-manager/client/src/ts/components/TagsSelector.tsx
```diff
@@ -0,0 +1,71 @@
...
+            return;
+        }
+
+        const newOption = {
+            label: searchValue,
+        };
+
+        // Create the option if it doesn't exist.
+        if (
+            flattenedOptions.findIndex(
```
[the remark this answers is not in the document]
rhystmills: (Now done)

apps/rule-manager/client/src/ts/components/RulesTable.tsx
```diff
@@ -122,16 +126,25 @@ const RulesTable = () => {
...
+                        items={rules}
+                        columns={columns}
+                        pagination={true}
+                        sorting={sorting}
+                        search={search}
+                    />
+                }
+            </EuiFlexItem>
+            <EuiFlexItem grow={1}>
+                <RuleForm fetchRules={fetchRules}/>
```
jonathonherbert, line 144: Non-blocking – wonder if this is better expressed as `onCreate` or `onChange`, that is to say, our RuleForm shouldn't know about fetching rules, but should provide a hook to do something when users click 'create' (or, later on, 'update')

apps/rule-manager/client/src/ts/components/TagsSelector.tsx
```diff
@@ -0,0 +1,65 @@
+import {useEffect, useState} from "react";
+import {EuiFormRow, EuiComboBox} from "@elastic/eui";
+import React from "react";
+import { MetadataOption } from "./CategorySelector";
+import { PartiallyUpdateRuleData, RuleFormData } from "./RuleForm";
+
+const existingTags = [
```
jonathonherbert, line 7: Non-blocking – should this live in 'constants'?

jonathonherbert (approved):
LGTM! A couple non-blocking comments.

apps/rule-manager/client/src/ts/components/TagsSelector.tsx
rhystmills: Good idea, now done!

apps/rule-manager/client/src/ts/components/RulesTable.tsx
rhystmills: I've gone with `onRuleUpdate` for now as that's specifically when it's called by the RuleForm currently

jonathonherbert (approved):

PR #319, opened 2023-05-02
Add update rule form (first pass)

rhystmills:
## What does this change?

This PR add an 'update rule' form to the rule manager.
As part of this work, we've also added a GET endpoint for an individual rule via its ID.

## How to test

With Composer credentials, run typerighter locally using `./script/start`. Click on the pencil icon next to a rule. Can you edit the rule via the form as intended?

## Accessibility

-   [X] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)


apps/rule-manager/client/src/ts/components/helpers/getRule.ts
```diff
@@ -0,0 +1,6 @@
+export const getRule = async (ruleId: number) => {
+    const getRuleResponse = fetch(`${location}rules/${ruleId}`, {
+         method: 'GET',
+     });
+     return getRuleResponse
```
jonathonherbert, line 5: Non-blocking: worth taking care of the status checking here? This fn's quite thin and reusing it across components would require the same checks each time – wonder if the same code could take care of get, create and update.

[Common types](https://github.com/guardian/editorial-tools-user-telemetry-service/blob/467caf5a96b9ef8a72f6f3cbf408df13a4d081d0/projects/event-lambdas/src/event-s3-lambda/index.ts#L25) for this can be useful, although `Either` is somewhat wishful thinking in that example, it's not a good match for the Scala equivalent 😀

apps/rule-manager/client/src/ts/components/helpers/updateRule.ts
```diff
@@ -0,0 +1,14 @@
+import { FormDataForApiEndpoint } from "./createRule";
+
+export const updateRule = async (ruleForm: FormDataForApiEndpoint) => {
+    // We would always expect the ruleForm to include an ID when updating a rule
+    if (!ruleForm.id) return (new Response(null, {status: 400, statusText: "Update endpoint requires a rule ID"}))
```
jonathonherbert, line 5: Related to the comment above about status checking, I don't think the interface for this should be a `Response` – coercing a synchronous validation error into that shape w/ a status code is a bit odd, it might make someone think a server was involved.

jonathonherbert (commented):
Pleasing! A few comments. Might be worth merging `main` into the target branch, and the target branch into this branch, to clean up the diff, as it's quite noisy ATM.

One thing to note – it's probably not for this PR, but we should think about how we're managing our routing. It'd be great to deep link to individual rules on the client, for example `/rule/1`.

That'll require us to namespace our api and frontend routes, so e.g. `/api/*` goes to the server, and `/app/` redirects to `/`, where the client can deal with the route. I'll make a follow up card.

apps/rule-manager/client/src/ts/components/helpers/getRule.ts
rhystmills: I've added a helper function to transform the responses from our API calls into our own type representing error response versus a success response.

apps/rule-manager/client/src/ts/components/RuleForm.tsx
```diff
@@ -66,13 +80,44 @@ export const RuleForm = ({fetchRules}: {fetchRules: () => Promise<void>}) => {
...
+                        setRuleData(baseForm);
+                        fetchRules();
+                        setCreateRuleFormOpen(false);
+                    } else {
+                        setErrors([...errors, {id: `${data.status} error`, value: `${data.statusText} - try again or contact the Editorial Tools team.`}])
+                    }
+                })
+        } else {
+            // Create new rule rather than updating existing rule
+            createRule(transformRuleFormData(ruleData))
```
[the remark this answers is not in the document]
rhystmills: Good shout, there's one path now.

apps/rule-manager/client/src/ts/components/RuleForm.tsx
```diff
@@ -66,13 +80,44 @@ export const RuleForm = ({fetchRules}: {fetchRules: () => Promise<void>}) => {
...
+                    }
+                })
+                .then(data => {
+                    if (data.status === 200) {
+                        setRuleData(baseForm);
+                        fetchRules();
+                    } else {
+                        setErrors([...errors, {id: `${data.status} error`, value: `${data.statusText} - try again or contact the Editorial Tools team.`}])
+                    }
+                })
```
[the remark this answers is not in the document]
rhystmills: I'd be keen to do a follow-up PR as this PR is quite chunky already, but definitely worth doing. I think a toast would work.

jonathonherbert (commented):
Looks good! Just one more thing I've found when updating rules I've just saved, or editing imported rules, that don't have tags:

```
dateRule.ts:16 Uncaught (in promise) TypeError: ruleForm.tags.join is not a function
    at transformRuleFormData (createRule.ts:17:62)
    at updateRule (updateRule.ts:6:28)
    at saveRuleHandler (RuleForm.tsx:84:22)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:3945:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
    at invokeGuardedCallback (react-dom.development.js:4056:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4070:25)
    at executeDispatch (react-dom.development.js:8243:3)
    at processDispatchQueueItemsInOrder (react-dom.development.js:8275:7)
    at processDispatchQueue (react-dom.development.js:8288:5)
```

Should be straightforward to reproduce, but lmk if not!

jonathonherbert (approved):

PR #320, opened 2023-05-04
Add feature switches, persisted w/ localStorage

jonathonherbert:
## What does this change?

Adds client-side feature switches. They're toggled with a context menu in the page header. They update the UI in real time.

I've added the username to the page header 'cos it seemed like the right thing to do 😀 

`Context` and localStorage do the state management. We're fairly comfortable that keeping switches on the client is sufficient at present.

feature-switches

## How to test

Have a play. Behave as expected?

NB: there's currently no way to write tests for the client – that'll come in another PR.

We can ship new features on the client without immediately exposing them to every user.


.tmp/cache/machine.json
aracho1, line 1: were these .tmp files meant to be git ignored?

aracho1 (approved):
Love the UI, tested on CODE and works great 💯

.tmp/cache/machine.json
jonathonherbert: Thanks for spotting! Fixed w/ a rebase. Files are generated by localstack and so not worried about history – I've added a gitignore entry for that path, too.

PR #323, opened 2023-05-05
Add 'rules_live' table; rename existing table to 'rules_draft'

Fweddi:


rhystmills (approved):
Code looks good, tested locally and apps are running as expected. The rule count reported by the main branch in the checker is the same as this branch 👍

PR #324, opened 2023-05-10
Add 'publish' action to backend

jonathonherbert:
## What does this change?

Adds a 'publish rule' action to the backend. This does two things:

- copies the rule from draft to live
- publishes a new rule artefact (which only includes live rules)

It's accessible via an endpoint, `/rules/:id/publish`. The publish action requires one addition param in the body, `reason`.

## How to test

- Try hitting that endpoint with the ID of a rule you've created in the UI. (You should be able to discover the ID via the network requests.) It should be published to the list of available rules in the checker. Other rules created but not published should not appear.


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -12,11 +12,14 @@ import com.gu.typerighter.model.{
...
 }
+import com.gu.typerighter.rules.BucketRuleResource
 import db.{DbRuleDraft, DbRuleLive}
 import db.DbRuleDraft.autoSession
 import scalikejdbc.DBSession
 
-object DbRuleManager extends Loggable {
+import scala.util.Try
+
+object RuleManager extends Loggable {
```
jonathonherbert, line 22: Renamed to reflect the fact that this manages rules in general, not just DbRules, including handling publishing duties. I moved publishing logic in here to get it out of the controller and into somewhere more testable.

apps/rule-manager/conf/routes
```diff
@@ -17,8 +17,11 @@ GET     /rules                      controllers.RulesController.rules()
 +nocsrf
 POST    /rules                      controllers.RulesController.create()
 +nocsrf
-GET     /rules/:id                  controllers.RulesController.rule(id: Int)
+GET     /rules/:id                  controllers.RulesController.get(id: Int)
```
jonathonherbert, line 20: Cleaning up a minor inconsistency in keeping our REST-ish methods verbs.

apps/rule-manager/test/db/DBTest.scala
```diff
@@ -35,5 +36,6 @@ trait DBTest extends BeforeAndAfter { self: Suite =>
 
   after {
     Evolutions.cleanupEvolutions(playDb)
+    scalikejdbcDb.closeAll
```
jonathonherbert, line 39: There were sporadic problems in tests where we'd get 'too many connections' from Postgres, I think because we weren't dropping connections. This seems to have fixed.

rhystmills (commented):
I've had a couple of odd responses from the API testing this, attempting to publish rule 13668 I got this rule (rule 1) back as a response:

image

I've also found that I got a generic 'Failed to add rule to Checker' error when I tried to publish a rule that didn't have a Category defined in the frontend edit form.

jonathonherbert:
@rhystmills thanks for spotting the problem with the odd id – this was a very odd bug caused by using `updateAndReturnGeneratedKey` without specifying the column name, and should be fixed in ([628a5e2](https://github.com/guardian/typerighter/pull/324/commits/628a5e28afc4e859e8e99797cebf2a293952029b).

I've also added form-based validation for checker rules ([8626cb3](https://github.com/guardian/typerighter/pull/324/commits/8626cb334e884d36cd2df92863fbce04c66fd308)...[d1e27af](https://github.com/guardian/typerighter/pull/324/commits/d1e27afa5079115cb44884afc3dffb854ab3f375)),  – the error you were seeing should definitely be there, but the format is very obtuse. Form-based validation now gives responses like:

```
{
    "errors": [
        {
            "key": "category",
            "message": "required"
        },
        {
            "key": "description",
            "message": "required"
        }
    ]
}
```

rhystmills:
I've tested again and I'm getting a `400` error back with the following message for a newly created rule that definitely has a description specified:
```
{"errors":[{"key":"description","message":"required"}]}
```
If I run a second time, I get a different error (`500`) with the following page, which makes me think the rule actually was published the first time:
image

rhystmills:
> I've tested again and I'm getting a `400` error back with the following message for a newly created rule that definitely has a description specified:
> 
> ```
> {"errors":[{"key":"description","message":"required"}]}
> ```
> 
> If I run a second time, I get a different error (`500`) with the following page, which makes me think the rule actually was published the first time: image
> 
> If the description is specified, but the category is missing, I only get the missing category as an error. If both are missing, I get both as an error.

(A rebase sorted this)

PR #325, opened 2023-05-11
Improve Vite behaviour when launching app with scripts

jonathonherbert:
## What does this change?

A few things to improve Vite behaviour when launching app with scripts:

- Stops vite removing output from previous commands from the terminal
- Makes sure the vite process ends when we stop running the app

## How to test

- Run the `start` script in the rule-manager client. It should be clear the output from Vite is not cleared.
- Harder to prove the second thing – perhaps try terminating the `start-manager` script a few times and check that vite has been properly cleared up? (The problem is intermittent, I think.)

### DIFF
--- a/apps/rule-manager/client/package.json
+++ b/apps/rule-manager/client/package.json
@@ -5,7 +5,7 @@
   "description": "",
   "type": "module",
   "scripts": {
-    "start": "vite",
+    "start": "vite --clearScreen false",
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
Existing build output is being retained when Vite launches and I've not been able to replicate the existing problem with Vite continuing to run after closing 👍

PR #327, opened 2023-05-15
Check for permissions before creating or editing rules on backend

rhystmills:
## What does this change?

1. Check for create and edit permissions on the backend before accepting a users request.
2. Only allow users to request rule creation or edit changes on the frontend if they have correct permissions.

Kapture 2023-05-15 at 10 56 40

## How to test

1. Run Typerighter locally according to the instructions in the readme.
2. Visit [manager.typerighter.local.dev-gutools.co.uk](https://manager.typerighter.local.dev-gutools.co.uk/)
3. Make sure you have 'Manage rules' Typerighter permissions [on permissions CODE](https://permissions.code.dev-gutools.co.uk/admin/user/111318226344948727644). Wait a minute for them to apply. Can you create and edit rules in the Typerighter manager locally?
4. Remove the permissions and wait a minute or so for the changes to persist. Can you no longer create and edit rules locally in Typerighter manager?

### DIFF
--- a/apps/rule-manager/client/src/ts/components/RuleForm.tsx
+++ b/apps/rule-manager/client/src/ts/components/RuleForm.tsx
@@ -6,8 +6,7 @@ import { RuleMetadata } from "./RuleMetadata";
 import { createRule } from "./api/createRule";
 import { FeatureSwitchesContext } from "./context/featureSwitches";
 import { updateRule } from "./api/updateRule";
-import { PageContext } from "../utils/window";
-import { hasCreateEditPermissions } from "./helpers/hasCreateEditPermissions";
+import { useCreateEditPermissions } from "./RulesTable";
 
 export type RuleType = 'regex' | 'languageToolXML';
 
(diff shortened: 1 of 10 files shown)


apps/rule-manager/client/src/ts/components/RuleForm.tsx
```diff
@@ -45,6 +45,7 @@ export const RuleForm = ({onRuleUpdate, ruleData, setRuleData, createRuleFormOpe
     }) => {
     const [showErrors, setShowErrors] = useState(false);
     const [errors, setErrors] = useState<FormError[]>([]);
+    const pageData = useContext(PageContext)
```
jonathonherbert, line 48: Non-blocking: just thinking that we do this in two places – we could make this slightly more concise by writing a `useCreateEditPermissions` hook that looks a bit like:

```ts
const useCreateEditPermissions = () => {
  const pageData = useContext(PageContext);
  // Do not recalculate permissions if the permissions list has not changed
  return useMemo(() => hasCreateEditPermissions(pageData.permissions), [pageData.permissions]);
}
```

(The memoisation optimisation is not super important, but arguably a good habit when calling pure fns from React components.)

apps/rule-manager/client/src/ts/components/RulesTable.tsx
```diff
@@ -104,7 +126,8 @@ const RulesTable = () => {
                 }
             })
     }
-    const columns = createColumns(openEditRulePanel);
+    const pageData = useContext(PageContext);
+    const columns = createColumns(openEditRulePanel, pageData ? hasCreateEditPermissions(pageData.permissions) : false);
```
jonathonherbert, line 130: Non-blocking: you could move the permissions hooks within `createColumns`, and save passing an argument here.

jonathonherbert (approved):
Tested client and server behaviour, works as expected. Reckon tooltips are a nice solution here, really clear! 🎉

apps/rule-manager/client/src/ts/components/RulesTable.tsx
rhystmills: Now done!

apps/rule-manager/client/src/ts/components/RuleForm.tsx
rhystmills: Now done.

jonathonherbert (approved):

PR #334, opened 2023-06-08
Add published status to rules table

jonathonherbert:
## What does this change?

Adds a 'status' column to the rule table, showing the publication status of a rule. (It'll also, eventually, show archived status.)

The UUID column is removed to make room for it, which satisfies the user story `I want to not have to worry about secondary rule information (ie. ID)`. You can still search by UUID.

Screenshot 2023-06-07 at 08 59 40

@Fweddi if this has a painful overlap with your 'archived' feature, let's chat, it might be simpler for me to rebase this on top of your work!

## How to test

- The automated tests should pass, and give us some confidence the query is correct.
- Load rules from the sheet into the management system with 'refresh rules'. They should all be live.
- Create a new new rule. It should be marked draft.
- Publish the rule, following the instructions in #324. It should now be marked live.

### DIFF
--- a/apps/rule-manager/client/src/ts/components/RulesTable.tsx
+++ b/apps/rule-manager/client/src/ts/components/RulesTable.tsx
@@ -60,10 +60,6 @@ const createColumns = (editRule: (ruleId: number) => void): Array<EuiBasicTableC
       name: 'Type',
     },
     {
-      field: 'externalId',
-      name: 'ID'
-    },
-    {
       field: 'category',
       name: 'Category',
     },
@@ -78,6 +74,11 @@ const createColumns = (editRule: (ruleId: number) => void): Array<EuiBasicTableC
     {
       field: 'description',
       name: 'Description'
+    },
+    {
+      field: 'isPublished',
+      name: 'Status',
+      render: value =>  value ? "Live" : "Draft"
     },
     {
       name: <EuiIcon type="pencil"/>,
--- a/apps/rule-manager/test/db/RuleManagerSpec.scala
+++ b/apps/rule-manager/test/db/RuleManagerSpec.scala
@@ -208,7 +208,9 @@ class RuleManagerSpec extends FixtureAnyFlatSpec with Matchers with AutoRollback
 
       RuleManager.destructivelyPublishRules(allRules, bucketRuleResource)
 
-      DbRuleDraft.findAll() shouldMatchTo allRules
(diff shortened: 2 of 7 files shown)


apps/rule-manager/client/src/ts/components/RulesTable.tsx
```diff
@@ -59,10 +59,6 @@ const createColumns = (editRule: (ruleId: number) => void): Array<EuiBasicTableC
       field: 'ruleType',
       name: 'Type',
     },
-    {
-      field: 'externalId',
-      name: 'ID'
-    },
```
simonbyford, line 65: Out of interest, why are we removing this column?

jonathonherbert: Ooh, should've mentioned this in the description – it makes room for the 'status' column in our table, and also fits our user story `✅I want to not have to worry about secondary rule information (ie. ID)'` (story doc [here.](https://docs.google.com/document/d/1dPSplzFR8RkQyPP0TSXlGkahmFU1q-znKneflursWVs/edit)) Description updated – hope that makes sense!

simonbyford: Cool, makes sense! 👌

apps/rule-manager/test/db/RuleManagerSpec.scala
```diff
@@ -208,7 +208,9 @@ class RuleManagerSpec extends FixtureAnyFlatSpec with Matchers with AutoRollback
 
       RuleManager.destructivelyPublishRules(allRules, bucketRuleResource)
 
-      DbRuleDraft.findAll() shouldMatchTo allRules
+      DbRuleDraft.findAll() shouldMatchTo allRules.map(rule =>
+        rule.copy(isPublished = !rule.ignore)
```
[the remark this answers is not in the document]
jonathonherbert: The rules we've created on L204 should be published when `ignore` is false when we call `destructivelyPublishRules`. This is our one-off spreadsheet import operation.

So the fact that `isPublished = false` when `rule.ignore` = false is really a side-effect of that process, based on the one-time import. It's possible in the future that these values could disagree.

We could just drop the `ignore` column. Its value is that it can tell us something historical about an imported rule – but ignored rules are not used, and I think it's somewhat unlikely we will care about them in the future.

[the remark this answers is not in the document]
aracho1: I see, thanks for the note 👍

PR #335, opened 2023-06-09
Add publish button with server-side validation

jonathonherbert:
## What does this change?

Adds a publish button with server-side validation. When a rule is changed, we ping the server to ask if the rule is publishable. If it's not, we get a tooltip with a list of reasons.

publish-validate

## How to test

- The automated tests should pass. This PR fixes a fairly egregious bug that the UI made obvious – publishing live rules affected the active status of every other live rule! Addressed in [ca0281b](https://github.com/guardian/typerighter/pull/335/commits/ca0281bed36af3c8234a8c2acc84818702ac58b8) with accompanying test.
- Try adding and publishing a few rules. It should work as expected – validation should work, and publishing valid rules should update the rule list and the rule history.

(Editing rules is a bit of a pain, now, in that when you hit 'update', you automatically navigate away from the current rule, and must re-open it to publish. Best to address in another PR, possibly by implementing real-time debounced updates.)

### DIFF
--- a/apps/rule-manager/client/src/ts/components/hooks/useRule.ts
+++ b/apps/rule-manager/client/src/ts/components/hooks/useRule.ts
@@ -44,11 +44,15 @@ export type RuleData = {
 
 export function useRule(ruleId: number | undefined) {
   const [isLoading, setIsLoading] = useState(false);
+  const [isPublishing, setIsPublishing] = useState(false);
+  const [isValidating, setIsValidating] = useState(false);
+  const [publishingErrors, setPublishingErrors] = useState<Error[] | undefined>(undefined);
   const [errors, setErrors] = useState<Error[] | undefined>(undefined);
   const [rule, setRule] = useState<RuleData | undefined>(undefined);
 
   const fetchRule = async (ruleId: number) => {
     setIsLoading(true);
+    setIsValidating(true); // Mark the rule as pending validation until the server tells us otherwise
 
     try {
       const response = await fetch(`${location}rules/${ruleId}`);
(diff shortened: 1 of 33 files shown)


aracho1:
One minor bug. After I publish a rule, and try updating the rule in the same form (because the form does not close or refresh automatically after publishing), it gives me an error (but there is no error message). 
image

To prevent this we could force the form to close after publishing (as I've observed that re-opening the form and updating the rule works as expected).

jonathonherbert:
@aracho1 thanks for spotting this! It required a bit of a refactor to resolve, but it should be solved in https://github.com/guardian/typerighter/pull/339 – would it make sense to merge this and fix-forward in that PR?

Fweddi (commented):
On creating a draft rule, we surface the following error when hovering over the Publish button:

image

It might be better to surface that the rule needs to be saved before it needs to be published. But, as I'm writing this, I recall that we are auto-saving to draft in another outstanding PR - so I think we can let this through.

jonathonherbert:
@Fweddi agreed that we _could_ leave this for the auto-saving story – I'm struggling to replicate, mind! Could you walk me through it?

apps/rule-manager/client/src/ts/components/hooks/useRule.ts
```diff
@@ -67,6 +71,48 @@ export function useRule(ruleId: number | undefined) {
...
+  }
+
+  const validateRule = async (ruleId: number) => {
+    setIsValidating(false);
+
+    try {
+      const response = await fetch(`${location}rules/${ruleId}/publish`);
+      if (response.status === 400) {
+        const validationErrors: Error[] = await response.json();
+        return setPublishingErrors(validationErrors);
```
Fweddi, line 106: It looks like we're setting 'publishing errors' when validating, and setting just 'errors' when publishing. These states might be worth renaming for clarity.

Fweddi:
> @Fweddi agreed that we _could_ leave this for the auto-saving story – I'm struggling to replicate, mind! Could you walk me through it?

I think what is happening is that when you hit the create button, it doesn't call the validation endpoint, and so retains the errors from the previous rule. In my case I was looking at an invalid live rule, and then hitting create.

The validation endpoint is only called when `rule` is truthy: https://github.com/guardian/typerighter/blob/1c1a847e5cabe06b710a7eeac403568fc10e552a/apps/rule-manager/client/src/ts/components/RuleForm.tsx#L59-L66

But when you hit the create button, you set the rule ID as undefined: https://github.com/guardian/typerighter/blob/1c1a847e5cabe06b710a7eeac403568fc10e552a/apps/rule-manager/client/src/ts/components/RulesTable.tsx#L210-L213
https://github.com/guardian/typerighter/blob/1c1a847e5cabe06b710a7eeac403568fc10e552a/apps/rule-manager/client/src/ts/components/RulesTable.tsx#L139-L142

Which in turn sets the rule as undefined:
https://github.com/guardian/typerighter/blob/1c1a847e5cabe06b710a7eeac403568fc10e552a/apps/rule-manager/client/src/ts/components/hooks/useRule.ts#L116-L122

Which means `rule` is falsy and we don't validate when hitting the create button.

jonathonherbert:
Thanks for spotting! [8f8a92a](https://github.com/guardian/typerighter/pull/335/commits/8f8a92abb9bdffab8b4aeed4624a50d8a32ae1be) should fix.

apps/rule-manager/client/src/ts/components/hooks/useRule.ts
jonathonherbert: Renamed to `publishValidationErrors` in c0484b313f05a124793fb0e3cd8251f078b987c1.

Fweddi (approved):
LGTM!

PR #337, opened 2023-06-13
Fix TS errors

jonathonherbert:
## What does this change?

Our TS build was not typechecking errors! 😱 This PR adds typechecking, and fixes the errors that exist in the current build.

It bumps TS for good measure.

## How to test

Try adding and editing rules on the client. Does everything work as expected?

### DIFF
--- a/apps/rule-manager/client/src/ts/utils/error.ts
+++ b/apps/rule-manager/client/src/ts/utils/error.ts
@@ -1,0 +1,2 @@
+export const errorToString = (error: unknown): string =>
+  error instanceof Error ? error.message : error ? error.toString() : typeof error;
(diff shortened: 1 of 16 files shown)


apps/rule-manager/client/src/ts/utils/error.ts
```diff
@@ -0,0 +1,2 @@
+export const errorToString = (error: unknown): string =>
```
jonathonherbert, line 1: Necessary b/c newer versions of Typescript types errors as `unknown`.

apps/rule-manager/client/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "jsx": "react-jsx"
   },
   "include": [
-    "react/src"
+    "src"
```
jonathonherbert, line 20: The mistake that caused typechecking to be skipped in the first instance.

aracho1 (approved):
Tested and LGTM 👍

PR #338, opened 2023-06-14
Add API routes to manage tags

rhystmills:
## What does this change?

This PR adds the following API routes for managing Tags:
- `GET     /tags`
- `POST    /tags`
- `GET     /tags/:id`
- `POST    /tags/:id`
- `DELETE  /tags/:id`

Following PRs will support managing the relationships between rules and tags.

## How to test

Run the rule manager locally (`.script/start`) and hit the above endpoints, e.g. https://manager.typerighter.local.dev-gutools.co.uk/tags

Can you:
- Retrieve the list of tags
- Add a new tag
- Get a specific tag by id
- Modify a tag's `name`
- Delete a tag by id

### DIFF
--- a/apps/rule-manager/app/controllers/TagsController.scala
+++ b/apps/rule-manager/app/controllers/TagsController.scala
@@ -3,7 +3,7 @@ package controllers
 import com.gu.pandomainauth.PublicSettings
 import com.gu.permissions.PermissionDefinition
 import com.gu.typerighter.lib.PandaAuthentication
-import model.CreateTagForm
+import model.{TagForm}
 import play.api.libs.json.Json
 import play.api.mvc._
 import utils.{FormHelpers, PermissionsHandler, RuleManagerConfig}
@@ -32,12 +32,22 @@ class TagsController(
         Ok(Json.toJson(tag))
     }
   }
+
+  def delete(id: Int) = ApiAuthAction {
+    Tags.find(id) match {
+      case None => NotFound("Tag not found matching ID")
+      case Some(tag) =>
+        Tags.destroy(tag)
+        Ok(s"Tag with id ${tag.id.get} deleted.")
+    }
+  }
+
   def create = ApiAuthAction { implicit request =>
     {
       hasPermission(request.user, PermissionDefinition("manage_rules", "typerighter")) match {
         case false => Unauthorized("You don't have permission to create tags")
         case true =>
-          CreateTagForm.form
+          TagForm.form
             .bindFromRequest()
             .fold(
               formWithErrors => {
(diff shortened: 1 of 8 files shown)


apps/rule-manager/app/db/Tags.scala
```diff
@@ -74,6 +79,36 @@ object Tags extends SQLSyntaxSupport[Tag] {
...
+      .toRight(NotFound("Rule not found matching ID"))
+      .map(existingRule =>
+        existingRule.copy(
+          name = tagForm.name
+        )
+      )
+    tag match {
+      case Right(tag) => {
+        Tags.save(tag).toEither match {
+          case Left(e: Throwable) => Left(InternalServerError(e.getMessage))
```
jonathonherbert, line 104: `InternalServerError` is a controller-level concept – our DB shouldn't care about HTTP things? Can we pass the `Either` back to the controller, and deal with the appropriate HTTP responses there?

jonathonherbert (approved):
Tested locally, routes looking good! One minor comment about error handling.

apps/rule-manager/app/db/Tags.scala
rhystmills: In this case I wanted to keep the output of this function to be `Either[Result, Tag]`, basically to identify errors where they're generated and continue operations on the `Right` from that point forward. e.g. It was helpful to identify the `NotFound` above when we know a rule hasn't been found. 

Perhaps we could go with `Either[Either[Result, string], Tag]` where the string is generic error messages to become `InternalServerError` down the line?

jonathonherbert: I see, you'd like to distinguish between 'not found' and 'something went wrong', and respond w/ the corresponding status codes. That's a good idea.

It's really easy to extend exceptions for specific circumstances – something like `class NotFoundException extends Exception` would give us a unique type that we could match on in the controller to produce the right result? (We could extend that outwards to other DB methods, although that's probably not for this PR.)

Could add it to e.g. `Errors.scala` in the utils package if that sounded good.

While I'm looking at this code, is there a reason that we're getting the rule and modifying it, before we update it? Perhaps to give the right 404? The update should fail if the tag doesn't exist, but perhaps the error we receive from the DB is not specific enough to be able to translate predictably?

rhystmills: Ah, custom error types is a nice idea, I'll give that a go.

That was the idea - the type for `Try` `toEither()` is the rather vague `Either[Throwable, T]`. It does seem a bit indirect, so it seems like the point you can figure out what kind of error it is is the point at which it's generated.

rhystmills: Now done!

apps/rule-manager/app/controllers/TagsController.scala
```diff
@@ -0,0 +1,90 @@
...
+          .fold(
+            formWithErrors => {
+              val errors = formWithErrors.errors
+              BadRequest(Json.toJson(errors))
+            },
+            tagForm => {
+              Tags.updateFromTagForm(id, tagForm) match {
+                case Left(NotFoundException(message)) => NotFound(message)
+                case Left(DbException(message))       => InternalServerError(message)
+                case Left(e: Exception)               => InternalServerError(e.getMessage)
```
jonathonherbert, line 83: Non-blocking – I think we can merge this and L82?

jonathonherbert (approved):

PR #339, opened 2023-06-14
Spruce page and table layout

jonathonherbert:
## What does this change?

Makes a few changes to the page layout and sidebar workflow to make the process of creating and editing rules a bit smoother:

- The table is now full-width, per the designs. (I haven't added the 90px gutter in this PR.)
- The table columns and widths should match the designs, and truncate text where necessary.
- The edit form disappears when rules aren't edited.
- Updating a rule no longer closes the form, to allow the user to update a rule, and then immediately publish it.
- Attempting to navigate away from the form requires confirmation from the user when the form has been edited.
- I've renamed the the 'discard changes' button to 'close', as the warning should  prevent lost work, and work will not be discarded every time that button is pressed (for example, if a user is browsing rules, rather than editing.)

|Form state|Before|After|
|--|--|--|
|Closed|Screenshot 2023-06-14 at 12 17 06|Screenshot 2023-06-14 at 10 37 30|
|Open|Screenshot 2023-06-14 at 12 17 28|Screenshot 2023-06-14 at 12 16 28|

## How to test

Have a play browsing, editing, and publishing rules. Does everything work as expected? Do the changes mark an improvement?


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -14,8 +14,21 @@ import db.{DbRuleDraft, DbRuleLive}
 import db.DbRuleDraft.autoSession
 import model.{LTRuleCoreForm, LTRuleXMLForm, RegexRuleForm}
 import play.api.data.FormError
+import play.api.libs.json.Json
 import scalikejdbc.DBSession
 
+object AllRuleData {
```
jonathonherbert, line 20: To correctly handle updating rule data client-side post-publication, it made sense to pass back all the rule data – and because we're now doing that in two places, a case class seemed better.

apps/rule-manager/client/src/ts/components/hooks/useRule.ts
```diff
@@ -116,6 +117,59 @@ export function useRule(ruleId: number | undefined) {
     }
   }
 
+  const updateRule = async (ruleForm: DraftRule) => {
```
jonathonherbert, line 120: Consolidating update and create operations in this effect lets them share error and loading states, which makes our UI more consistent (loading state now accounts for fetching, creating, and updating, for example.)

Fweddi (approved):
All LGTM!

Added a tiny tweak to improve the tag wrapping as tags were still overflowing the column width.

jonathonherbert:
Thanks @Fweddi!

PR #340, opened 2023-06-14
Add reason when republishing rule via modal

jonathonherbert:
## What does this change?

Adds a modal asking the user to provide a reason when republishing a rule. The reason appears in the rule history.

It's mandatory at present. Too much?

add-reason

## How to test

Go through the workflow shown in the GIF. Does everything work as expected?

### DIFF
--- a/apps/rule-manager/client/src/ts/components/RuleForm.tsx
+++ b/apps/rule-manager/client/src/ts/components/RuleForm.tsx
@@ -16,6 +16,8 @@ import {RuleHistory} from "./RuleHistory";
 import styled from "@emotion/styled";
 import {capitalize} from "lodash";
 
+import { ReasonModal } from "./modals/Reason";
+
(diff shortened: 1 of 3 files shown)


Fweddi (approved):
LGTM.

Perhaps it should be optional - but then the choice of a modal is a little intrusive.

I also wonder whether we should allow users to add a reason on first publish?

Also worth noting that currently if you republish an archived rule it asks for a reason (even though this is not a revision of content).

jonathonherbert:
@Fweddi the reasoning for a lack of modal on first publish – the description of the rule should be its justification as it goes out. But there may be context that's useful on first publish, e.g. 'Added in line with new style guidance, 26/06/15'.

Let's talk to Ana and our users a bit more about the 'reason' part of our work, I think this idea stands to be improved.

I think there's a follow up PR to be writtent to make `archive` and `publish` make sense to eachother, as we wrote each feature independently.

PR #341, opened 2023-06-15
Add checkboxes to select multiple rules

aracho1:
## What does this change?
Add checkboxes to the rules table to enable users to select multiple rows in preparation for the batch editing feature to come. 

Clicking the header checkbox will select all the rows in the current page. Selecting one or multiple rows will bring up the edit form. If multiple rules are selected, the details of the first selected rule will show up. The form does not do anything in relation to batch editing for now but we'll bring the relevant changes in subsequent PRs.

image

## How to test
Run this locally or on CODE and check that one more more rules can be selected. The edit form should show up upon selecting one or more rows.


jonathonherbert (approved):
Great work! Nice touch to make sure the form is open and closed correctly.

PR #342, opened 2023-06-15
Add permission to access permissions

jonathonherbert:
## What does this change?

#118 added permissions – but we're missing S3 access to fetch the permissions cache in CODE and PROD, [hence these logs.](https://logs.gutools.co.uk/s/editorial-tools/goto/627b41f0-0bc2-11ee-872e-07cde0a14815)

This PR adds access via the CDK.

## How to test

CODE should currently deny you permission to edit rules, regardless of whether they're set, as the box cannot load them. Deploy to CODE, granting yourself permission to add rules. You should be able to edit rules once the box refreshes its permissions – within a 1 minute window.

- [x] Tested in CODE

### DIFF
--- a/cdk/lib/index.ts
+++ b/cdk/lib/index.ts
@@ -75,6 +75,11 @@ export class Typerighter extends GuStack {
       bucketName: "pan-domain-auth-settings",
     });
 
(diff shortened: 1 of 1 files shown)


aracho1 (approved):
Tested in CODE 👍

PR #346, opened 2023-06-20
Batch edit rules (server-side)

aracho1:
## What does this change?

This PR introduces a new endpoint  `/rules/batch` for batch editing, and adds `batchUpdate` method to authenticate and validate (via `BatchUpdateRuleForm`) the request. Lastly `DbRuleDraft.batchUpdateFromFormRule()` updates the corresponding rules in the database. I have also added a test.

## How to test
Run this locally (or on CODE but #342 would need to be merged to fix CODE permissions). Create multiple new rules and and send a POST request to  `/rules/batch` with their IDs and new categories and/or tags to update e.g.
```
{
 "ids":[800,801,802], 
 "fields": {
    "category": "Check this",
    "tags": "General,SG"
  }
}
```
Check that the relevant fields have been updated for these rules.

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -4,7 +4,7 @@ import com.gu.pandomainauth.PublicSettings
 import com.gu.permissions.PermissionDefinition
 import com.gu.typerighter.lib.PandaAuthentication
 import com.gu.typerighter.rules.BucketRuleResource
-import play.api.libs.json.Json
+import play.api.libs.json.{JsSuccess, JsValue, Json, Reads}
 import db.DbRuleDraft
 import model.{CreateRuleForm, PublishRuleForm, UpdateRuleForm}
 import play.api.mvc._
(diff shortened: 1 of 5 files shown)


apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -270,6 +270,44 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
+          val updatedRows = withSQL {
+            update(DbRuleDraft)
+              .set(
+                column.category -> mergedRule.category,
+                column.tags -> mergedRule.tags,
+                column.updatedBy -> user,
+                column.revisionId -> sqls"${column.revisionId} + 1"
+              )
+              .where
+              .eq(column.id, id)
```
jonathonherbert, line 298: There's one drawback to this approach – we're making a database call to read and write for each rule we're updating. For larger numbers of rows, this will be very slow.

We should be able to write an update clause that uses an `in` condition to do this, something like:

```suggestion
              .in(column.id, listOfIdsToUpdate)
```

Then, we should be able to lose the `.map` in favour of a single DB call. I think the update call will fail if a rule does not exist, and that should be enough error handling for batch calls – I wouldn't worry about reading every rule to make sure it's there before we update. (We can always write a test to confirm, and if the error is very obscure we can consider making it more readable as we handle it.)

Full disclosure, although I'm confident this approach will work, I've never done this before with ScalikeJDBC, and it might be you've tried this and found problems with this approach – v. happy to pair!

jonathonherbert (commented):
Tested locally, works very well. Good work.

I've added a few recommendations that I think are worth implementing – hopefully they make sense!

apps/rule-manager/app/controllers/RulesController.scala
```diff
@@ -121,4 +121,78 @@ class RulesController(
...
+              },
+              batchUpdateRequest => {
+                val ids = batchUpdateRequest.ids
+                if (ids.isEmpty) BadRequest("No rule IDs provided")
+                else {
+                  val updatedRuleForm = UpdateRuleForm(
+                    category = batchUpdateRequest.fields.category,
+                    tags = batchUpdateRequest.fields.tags
+                  )
+                  val filledForm = UpdateRuleForm.form.fill(updatedRuleForm)
```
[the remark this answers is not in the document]
aracho1: If we pass categories and tags directly to the batch update method, then don't we have to pass category and tags in the query params i.e. `/routes/batch/:category/:tags`? My concern is then tags could end up being very long, that's why I thought it's better to put them in the request body. But this could be my incomplete Scala/Play understanding!

aracho1:
Thank you for your review and suggestions @jonathonherbert! I've tried to address them in my latest commit, appreciate if you could take another look. If anything looks wrong, please let me know.

apps/rule-manager/app/controllers/RulesController.scala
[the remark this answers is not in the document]
jonathonherbert: I think this comment above was directed at the arguments in the function (I cannot now see the history) – you're absolutely right, we wouldn't want to specify them in the route for a few reasons, one being URL length.

The updated code with the form looks great.

apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -276,6 +276,42 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
+            column.updatedBy -> user,
+            column.revisionId -> sqls"${column.revisionId} + 1"
+          )
+          .where
+          .in(column.id, ids)
+      }.update().apply()
+
+      if (updatedRows > 0) {
+        ids.flatMap { id =>
+          find(id) match {
```
jonathonherbert, line 297: The update clause looks perfect, but we're running into the same problem with this select clause – it'll run once for every rule, which on an update of thousands will be very slow.

Instead, could we try to select many rules by id? We could write a new version of `findAll`, perhaps, that accepted a list of ids, and used a `where.in(...)` clause to select the relevant rules. This'd be nice and quick.

This should mean that we don't have to worry about the rehydration of categories/tags on L301/2 (although I'm not quite sure w

aracho1: That's a good point, thanks for raising it 👍

jonathonherbert (approved):
Looks great!

PR #347, opened 2023-06-20
Allow users to archive a rule

Fweddi:
## What does this change?

Adds an archive button. When a rule is archived, we set the rule as archived in the draft table. If there's a corresponding active rule in the live table, we set that rule as inactive.

https://github.com/guardian/typerighter/assets/40991816/e2f4cc99-3740-4960-b152-9e36275a44b6

It is worth nothing that we are mutating live rules here (previously each new version of a rule has been inserted as a new row). We think is okay, as we are modifying the 'is_active' column, rather than the rule contents.

The UX is a first pass - awaiting final design.

## How to test

- Try archiving draft and live rules using the archive button (prior to https://github.com/guardian/typerighter/pull/335, you may need to make rules live using Postman / Postgres). The status in the Rules table should change correspondingly.

In future passes we hope to add the ability to filter out archived rules, the ability to unarchive a rule, and the ability to see the archive action in the publication history (this last part might be a little complex given we do not model archiving on the live rule itself).

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -121,4 +121,21 @@ class RulesController(
           )
     }
   }
+
+  def archive(id: Int): Action[AnyContent] = ApiAuthAction { implicit request =>
+    hasPermission(request.user, PermissionDefinition("manage_rules", "typerighter")) match {
+      case false => Unauthorized("You don't have permission to archive rules")
+      case true =>
+        RuleManager.archiveRule(id, request.user.email) match {
+          case Left(e: Throwable) => InternalServerError(e.getMessage)
+          case Right((draftRule, liveRule)) =>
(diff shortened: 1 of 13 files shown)


apps/rule-manager/app/controllers/RulesController.scala
```diff
@@ -121,4 +121,21 @@ class RulesController(
           )
     }
   }
+
+  def archive(id: Int): Action[AnyContent] = ApiAuthAction { implicit request =>
+    hasPermission(request.user, PermissionDefinition("manage_rules", "typerighter")) match {
```
jonathonherbert, line 126: Not for this PR, but a reminder to self that we should find an abstraction for this check, as it's very common!

jonathonherbert (approved):
Tested locally, works as expected. Code looks great. Quick, merge it before the publish work goes in 😀

PR #348, opened 2023-06-27
Rearchitect rule states and transitions

Fweddi:
## What does this change?

Rules can now move along the following states: `Archived` <-> `Draft` <-> `Live`

They can only move one state along at a time. For example, previously a user could archive from live. Now they must first unpublish a rule to draft, and then archive the draft rule.

The possible states and transitions can be summarised as follows:

| Rule State | Valid Transitions |
| --- | --- |
| Archived | Unarchive (-> Draft) |
| Draft | Archive (-> Archived), Publish (-> Live) |
| Live | Unpublish (-> Draft) |

This PR refactors our two existing state functions, Publish and Archive, to reflect the above, and adds two new state functions, Unarchive and Unpublish. 

We validate state transitions on the client and server, and show different actions to the user depending on the rule state (e.g. if the rule is archived, we only show the `Unarchive` button, and do not surface the `Publish` button).

## How to test

Run the new and updated tests - they should pass.

Play around with the various state buttons - you should be able to move a rule into different states without any errors.

A (relatively) simple publication system ready before launch, that allows users to manage rules throughout their life cycle.

### DIFF
--- a/apps/rule-manager/app/service/RuleManager.scala
+++ b/apps/rule-manager/app/service/RuleManager.scala
@@ -267,10 +267,10 @@ object RuleManager extends Loggable {
     }
   }
 
-  def archiveRule(
+  def unpublishRule(
       id: Int,
       user: String
-  ): Either[Exception, (Option[DbRuleDraft], Option[DbRuleLive])] = {
+  ): Either[Exception, Option[DbRuleLive]] = {
     try {
       val maybeDraftRule = DbRuleDraft.find(id)
 
@@ -282,13 +282,26 @@ object RuleManager extends Loggable {
         updatedLiveRule <- DbRuleLive.setInactive(externalId, user)
       } yield updatedLiveRule
 
+      Right(maybeUpdatedLiveRule)
+    } catch {
+      case e: Exception => Left(e)
+    }
+  }
+
+  def archiveRule(
+      id: Int,
+      user: String
+  ): Either[Exception, Option[DbRuleDraft]] = {
+    try {
+      val maybeDraftRule = DbRuleDraft.find(id)
+
       val maybeUpdatedDraftRule = for {
         draftRule <- maybeDraftRule
         archivedDraftRule = draftRule.copy(isArchived = true)
         updatedDraftRule <- DbRuleDraft.save(archivedDraftRule, user).toOption
       } yield updatedDraftRule
 
-      Right(maybeUpdatedDraftRule, maybeUpdatedLiveRule)
+      Right(maybeUpdatedDraftRule)
     } catch {
       case e: Exception => Left(e)
     }
--- a/apps/rule-manager/app/service/RuleManager.scala
+++ b/apps/rule-manager/app/service/RuleManager.scala
@@ -176,6 +176,18 @@ object RuleManager extends Loggable {
             )
           )
         )
+      _ <- draftRule match {
+        case _ if draftRule.isArchived =>
+          Left(
+            Seq(
+              FormError(
+                "Rule is archived",
+                "Only unarchived rules can be published"
+              )
+            )
+          )
+        case _ => Right(())
+      }
       liveRule = draftRule.toLive(reason)
       externalId <- liveRule.externalId.toRight(Seq(FormError("External id", "required")))
       _ <- liveDbRuleToCheckerRule(liveRule)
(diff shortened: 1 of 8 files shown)


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -176,6 +176,18 @@ object RuleManager extends Loggable {
...
+          Left(
+            Seq(
+              FormError(
+                "Rule is archived",
+                "Only unarchived rules can be published"
+              )
+            )
+          )
+        case _ => Right(())
+      }
```
Fweddi, line 190: The server side validation for publish is a little more complex than for the other state transitions, where we simply check for an if condition in a for comprehension.

apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -267,9 +279,10 @@ object RuleManager extends Loggable {
     }
   }
 
-  def archiveRule(
+  def unpublishRule(
       id: Int,
-      user: String
+      user: String,
+      bucketRuleResource: BucketRuleResource
   ): Either[Exception, (Option[DbRuleDraft], Option[DbRuleLive])] = {
```
Fweddi, line 286: There's an argument to be made that we could return `AllRuleData` here, in line with `publishRule`.

apps/rule-manager/client/src/ts/utils/rule.ts
```diff
@@ -1,9 +1,17 @@
 import {DraftRule} from "../components/hooks/useRule";
 import {IconColor} from "@elastic/eui/src/components/icon";
 
-type RuleState = "live" | "archived" | "draft";
+export type RuleState = "live" | "archived" | "draft" | "error";
```
Fweddi, line 4: Have added an extra type here on the client-side - "error" - to surface a buggy state more clearly. We could also choose to handle `RuleState | undefined`.

apps/rule-manager/app/service/RuleManager.scala
jonathonherbert: Non-blocking: IIUC we will probably need to do that in future if we'd like to improve our rule history by providing the end date of the published rule, as well as the start date.

(The end date will be the new information provided by `AllRuleData`.)

Could include that behaviour in this PR, although there's no UI I'm aware of that will surface the update at the moment.

apps/rule-manager/client/src/ts/components/RuleForm.tsx
```diff
@@ -148,6 +149,32 @@ export const RuleForm = ({ruleId, onClose, onUpdate}: {
...
 
+    const unarchiveRuleHandler = () => {
+        if (!ruleFormData?.id || ruleState !== 'archived') {
+            return;
+        }
+
+        unarchiveRule(ruleFormData.id).then(data => {
+            if (data.status === 'ok'){
+                setRuleFormData(baseForm);
+                onClose();
```
jonathonherbert, line 160: Do we really want to close the form on unarchive or archive? My feeling is that only clicking 'close' should close the sidebar, but you may feel differently!

apps/rule-manager/client/src/ts/components/hooks/useRule.ts
```diff
@@ -112,6 +114,30 @@ export function useRule(ruleId: number | undefined) {
...
+  const unarchiveRule = async (ruleId: number) => {
+    setIsLoading(true);
+
+    const result = await fetch(`${location}rules/${ruleId}/unarchive`, {
+      method: 'POST',
+    }).then(responseHandler)
+
+    setIsLoading(false);
+
+    return result;
```
jonathonherbert, line 126: At the moment, we're not updating the client representation of this rule (`[rule, setRule]`). We should do that – at the moment, we're closing the form, but (per another comment) I'm not convinced that we should, and if we don't close the form it'll be apparent that this state isn't being updated.

Per your suggestion, perhaps best to return `AllRuleData` and update `rule` in this way. It means we're always in sync with what the server thinks after our fetch completes.

jonathonherbert (commented):
What a lovely PR! Tested locally cycling through the different states, including checking the published bundle with (you can `cat` the output with `aws --endpoint-url=http://localhost:4566 s3 cp s3://typerighter-app-local/local/rules/typerighter-rules.json --profile composer -`.) All works as expected.

There's a blocking comment around updating client side state (YMMV on UI behaviour here, happy to discuss), but other than that this is looking great.

PR #350, opened 2023-06-27
Return 200 from GET publish endpoint

Fweddi:
## What does this change?

A small change to avoid polluting the console and log stack -> the GET /publish endpoint now returns a 200 rather than a 400.

## How to test

Errors from the endpoint populate the form as before. The network tab now shows only 200s for validation requests.

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -143,8 +143,8 @@ class RulesController(
 
   def canPublish(id: Int) = ApiAuthAction {
     RuleManager.parseDraftRuleForPublication(id, "validate") match {
-      case Right(_)     => Ok
-      case Left(errors) => BadRequest(Json.toJson(errors))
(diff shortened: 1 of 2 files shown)


apps/rule-manager/app/controllers/RulesController.scala
```diff
@@ -143,8 +143,8 @@ class RulesController(
 
   def canPublish(id: Int) = ApiAuthAction {
     RuleManager.parseDraftRuleForPublication(id, "validate") match {
-      case Right(_)     => Ok
-      case Left(errors) => BadRequest(Json.toJson(errors))
+      case Right(_)     => Ok(Json.toJson(Nil))
```
Fweddi, line 146: We need to return something here to make it easier to distinguish between these responses on the client-side (now the responses have merged in status code). You could argue for a different approach to what this distinctive marker is - I've opted for an empty list here.

jonathonherbert (approved):
Tested locally 👍  empty array is the way that I'd do it – 'q: any problems? a: 0 problems'.

PR #352, opened 2023-06-28
Better user experience when editing rules of different types

jonathonherbert:
## What does this change?

A few tweaks when editing rules of different types:

- the rule type is now respected, changing when the rule data changes
- the input changes from a text to textarea when editing XML rules, to reflect their multiline content

rule-type

I've also moved the radio buttons nearer the pattern input, as the two fields are very related – one affects the other's presentation.

## How to test

Have a play inspecting and editing rules of different types. Do the changes make sense?


rhystmills:
I think this also include changes from: https://github.com/guardian/typerighter/pull/354/files

Going to try rebasing on: https://github.com/guardian/typerighter/pull/354/files

rhystmills (approved):
Very sensible change and runs as expected for me.

Makes me realise that the UX for editing descriptions is quite rubbish given that they're often quite long. Would be good if they're multiline/resizeable in the future (either based on manual drag by user or automatically based on length):

image

PR #353, opened 2023-06-28
Highlight the rule that's currently being edited

jonathonherbert:
## What does this change?

Highlights the row of the currently edited rule.

It's a bit light by default in EUI, isn't it! Breaking all sorts of contrast rules, probably. What's a better shade? Perhaps the same colour as the form background?

Or perhaps we could get rid of the edit pencil altogether, and clicking anywhere on the row selects it, setting the checkbox on the left to ticked (a clearer signal?)

select-rule

## How to test

Have a play. Work as expected?


rhystmills (approved):
Definitely better than what we have now. Good suggestion with clicking a row to edit, does seem like the pencil doubles up on the checkbox functionality now.

apps/rule-manager/client/src/ts/components/RulesTable.tsx
```diff
@@ -238,6 +243,7 @@ const RulesTable = () => {
       <EuiFlexItem grow={2}>
         {rules &&
           <EuiInMemoryTable
+            rowProps={getRowProps}
             loading={isLoading}
```
Fweddi, line 247:
```suggestion
	css={css`.euiTableRow.euiTableRow-isSelected { background-color: rgba(0, 119, 204, 0.1); }`}
        loading={isLoading}
```

Fweddi: If we want a slightly more prominent selection colour

jonathonherbert: This is perfect to my eyes, nice!

PR #354, opened 2023-06-29
Autosave on rule edit

jonathonherbert:
## What does this change?

Adds autosave on rule edit for every field. Adds 'updated at' text that's refreshed on update, and periodically rerendered to ensure it doesn't get stale.

real-time-updates

Removes the 'update' button, which is no longer needed.

## How to test

Have a go at editing a rule. It should update as expected after a short delay, and the updated time should change as expected, even when not editing (currently it refreshes every 10 seconds, which felt about right, not too distracting!)

### DIFF
--- a/apps/rule-manager/client/src/ts/components/LastUpdated.tsx
+++ b/apps/rule-manager/client/src/ts/components/LastUpdated.tsx
@@ -1,0 +1,18 @@
+import { formatDistance } from "date-fns";
+import { usePeriodicRefresh } from "./hooks/usePeriodicRefresh";
+
+/**
+ * Return the
+ * @param lastUpdated
+ * @returns
+ */
+export const LastUpdated: React.FC<{ lastUpdated: string }> = ({
+  lastUpdated,
+}) => {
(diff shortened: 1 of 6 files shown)


rhystmills:
Works really well, feels intuitive to me, and code looks good.

I think one bit of functionality missing now that I try it is an equivalent of the discard functionality, or an undo. As an example user story:

1. Someone accidentally edits a live rule when their cat walks over their keyboard
2. The match is replaced and the rule is closed
3. They have no idea what the original match is, so have no way of restoring the current draft to its original state. 

Compounding that, it looks as if the changes are published to them, because the status says 'Live'

For example, a unpublished replacement:
image

For an unpublished draft the current setup is probably fine but it does feel like a reset should be possible when there's a published version. The '_has unpublished changes status'_ card would help (a bit) with the '_unpublished changes appearing to be live_' problem. Though better still would be some kind of '_unpublished change indicator_' alongside the specific fields that have been updated.

rhystmills (approved):

PR #355, opened 2023-06-30
Hide refresh button and make message scary

rhystmills:
## What does this change?

This hides the 'Refresh rules' button behind a feature switch, adding a scary new message with a scary red background ☠️

Collectively these two changes should make it very unlikely anyone will click the button accidentally.

We'll remove the button altogether once the tool is being used universally instead of the sheet.

## How to test

1. Run the manager locally according to the instructions in the readme.
2. Enabled the 'Enable destructive reload from rules sheet' feature switch in the top right dropdown menu
3. Can you refresh the rules via the scary button?

## Images

image


simonbyford (approved):
😨

PR #356, opened 2023-07-03
Batch edit rules (client-side)

aracho1:
## What does this change?
Enable batch edit on the client side to accompany server-side implementation (#346). This PR introduces a new rule form `RuleFormBatchEdit` and a new hook `useBatchRules` to allow users to edit multiple rules at the same time. Note that I've had to make changes to `DbRuleDraft.batchUpdate` to take into consideration requests where only category or tags are edited, not both.

With the current implementation, upon selecting multiple rules, the form will display the category of the first selected rule, and the tags section will show all the tags of the selected rules.
batch-edit

## How to test

1. Select multiple rules for testing using the tick boxes.
2. Update the category and/or the tags section and hit the update button. In the dev console, check that a POST request is made to `/rules/batch` endpoint with the updated fields as the payload. If either of category of tags was edited, the request should only include the relevant fields.
3. The new rules fetched in the table should now be updated with the new tags and/or categories.


apps/rule-manager/client/src/ts/components/RuleMetadata.tsx
```diff
@@ -8,14 +8,14 @@ import {DraftRule} from "./hooks/useRule";
...
+    ruleData: DraftRule | DraftRule[],
     partiallyUpdateRuleData: PartiallyUpdateRuleData,
     tags: TagMap
 }) => {
 
     return <RuleFormSection title="RULE METADATA">
         <LineBreak/>
-        <CategorySelector ruleData={ruleData} partiallyUpdateRuleData={partiallyUpdateRuleData}/>
-        <TagsSelector tags={tags} ruleData={ruleData} partiallyUpdateRuleData={partiallyUpdateRuleData}/>
+        <CategorySelector ruleData={Array.isArray(ruleData) ? ruleData[0]: ruleData} partiallyUpdateRuleData={partiallyUpdateRuleData}/>
```
aracho1, line 18: Not very happy with this, please do suggest if there is a cleaner way to implement this

apps/rule-manager/client/src/ts/components/hooks/useBatchRules.ts
```diff
@@ -0,0 +1,93 @@
...
+          'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(ruleFormData)
+    })
+
+    const parsedResponse = await responseHandler(response);
+    if (parsedResponse.status === "ok") {
+      if (rules  && parsedResponse.data) {
+        const updatedRules = rules.map((rule, index) => {
+          const updatedRule = (parsedResponse.data as unknown as DraftRule[]).find(updatedRule => rule.draft.id === updatedRule.id)
```
aracho1, line 70: Had to convert to `unknown` first as compiler thinks `parsedResponse.data` is a single `DraftRule` - open to feedback on if there is a cleaner way to achieve this

jonathonherbert:
Server side looks great! Spoken to @aracho1 and we're going to give this autosave in line with #354.

aracho1:
@jonathonherbert and I had further discussion and decided not to include auto-save in this PR

apps/rule-manager/client/src/ts/components/hooks/useBatchRules.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  const updateRules = async (ruleForm: DraftRule[]) => {
+    setIsLoading(true);
+
+    const hasCategoryChanged = rules && rules[0].draft.category !== ruleForm[0].category;
+
+    const haveTagsChanged = () => {
+        if (!rules) return false;
+        const oldTags = rules.map(rule => rule.draft.tags).flat()
```
jonathonherbert, line 36: Not blocking, but this (and other instances in this file) is equivalent to `flatMap` –
```suggestion
        const oldTags = rules.flatMap(rule => rule.draft.tags)
```

apps/rule-manager/client/src/ts/components/hooks/useBatchRules.ts
jonathonherbert: You're right, that's not ideal. Let's address this in a separate PR – `responseHandler` could do with a refactor.

jonathonherbert (approved):
This is a great start! Tested locally and in CODE, all works as expected.

I think as we discussed there are a few questions around batch editing semantics, but we can address in follow up PRs.

jonathonherbert:
(Hope you don't mind me rebasing from `main`, I'd like to chain #358 off this branch!)

apps/rule-manager/client/src/ts/components/hooks/useBatchRules.ts
aracho1: Good spot, thanks for that!

PR #358, opened 2023-07-03
Add rule state, and 'unpublished changes' status, to rule form.

jonathonherbert:
_co-authored-by: @Fweddi_ 

## What does this change?

Adds a 'rule state' section to the form, indicating whether a rule is draft, live, archived, as well as whether it has unpublished changes.

'Unpublished changes' means that the rule is published, and the draft rule has a different revision ID to the latest live rule. There's an edge case here – it means that someone may make and revert an edit, and the rule will still be marked as having unpublished changes. I think that's preferable to having to diff every 'contentful' field in a rule to figure this out (it's slower, and we must maintain that list as we add fields to our rules), but others may reasonably disagree!

To make room in the UI, we've consolidated the rule content and metadata parts of the form into a single section. This has the advantage of making the last updated time – `updated 3 minutes ago` – relate more clearly to the content of the rule (rather than e.g. its publication time.)

|Before|After|
|--|--|
|Screenshot 2023-07-03 at 18 09 15|Screenshot 2023-07-03 at 18 02 58|

## How to test

Running locally or in code, move a rule between different states – archived, draft, published. Does the status respond accordingly? Does the layout change make sense?


Fweddi (approved):
So good!

We paired on this together and I have double-checked it works locally.

In a future PR we hope to surface the 'unpublished changes' in the rule table, too.

PR #360, opened 2023-07-04
Rejig rule table

Fweddi:
## What does this change?

_Apologies for the one commit - Git wreaked havoc on my commit history, and I haven't the heart to go back over my commits and rebase._

image

This PR tweaks the order of the fields -> "Description", "Pattern", "Replacement" in the form, and in the table. We feel the order should be like this because:

- Description best describes a rule, and is always present
- Match is always present, and reading left to right from ‘match’ to ‘replacement’ makes sense
- Replacement is not always present

We have also renamed some fields, and added some handling for empty fields (here we now display 'em dash': –)

## How to test

Is the form and table more consistent? Do the changes make sense?

### DIFF
--- a/apps/rule-manager/client/src/ts/components/CategorySelector.tsx
+++ b/apps/rule-manager/client/src/ts/components/CategorySelector.tsx
@@ -25,7 +25,7 @@ export const CategorySelector = ({currentCategory, partiallyUpdateRuleData}: {
     }, [selectedCategory])
 
     return (
-        <EuiFormRow label='Category' fullWidth={true}>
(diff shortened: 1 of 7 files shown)


jonathonherbert (approved):
Looks good. Worth making our `–` 'subdued? At the moment, impossible to distinguish between a field that contains that character.

Fweddi:
> Looks good. Worth making our `–` 'subdued? At the moment, impossible to distinguish between a field that contains that character.

I tried it with `subdued` and `lightShade` - both were far too light to make out. Could try another colour if you want, but I'd like to think that it's very unlikely any of our fields will contain a single em dash!

PR #361, opened 2023-07-05
Description field with markdown preview

samanthagottlieb:
## What does this change?

This changes the rule Description field:
- makes it multiline, expandable and scrollable
- adds a Preview button to preview the markdown

@rhystmills and I explored adding an EuiMarkdownEditor, however this came with a toolbar that included plugins to add a task list, code snippet, etc. Because these are not supported by Composer, we thought it would be best not to show these options to the user. Unfortunately the only plugin EUI allows you to disable is 'tooltip'.

Following discussion in the Typerighter chat group, the team agreed that implementing a text box that supports markdown (to allow the user to make things bold or italic if they know the syntax) and allows you to preview how it will look in Composer is a good step in the right direction for now!

Trello card is [here](https://trello.com/c/BQmbEfDG/1522-rule-description-field-should-be-multiline-and-also-support-formatting).

## How to test

This has been tested by running the manager app locally.

### Before
before-description-preview

#### Write mode:
after-description-preview-hidden

#### Preview mode:
after-description-preview-shown


rhystmills (commented):
Looks great and works as expected.

I wonder if it would be worth adding some styles to differentiate the preview from the body of the form, though not sure exactly what the best option would be. Perhaps a lighter background and a border? Might be worth getting some UX input.

Also not sure if we should shout out that it's a markdown field in some way in the interface. Again, not sure the best way to do this, perhaps a tooltip (I think EUI has one).

samanthagottlieb:
> Looks great and works as expected.
> 
> I wonder if it would be worth adding some styles to differentiate the preview from the body of the form, though not sure exactly what the best option would be. Perhaps a lighter background and a border? Might be worth getting some UX input.
> 
> Also not sure if we should shout out that it's a markdown field in some way in the interface. Again, not sure the best way to do this, perhaps a tooltip (I think EUI has one).

Both great points, thanks @rhystmills. I've changed the background to match the colour and padding of other fields once the preview button is pressed. In Preview mode, the surrounding box will jump in/out to hug the text, with the same amount of padding as other input fields:

Screenshot 2023-07-06 at 17 47 20
Screenshot 2023-07-06 at 17 47 31

Some UX input is probably a good idea, I'll reach out.

I've also added a tooltip:

Screenshot 2023-07-06 at 17 48 06

rhystmills:
Looking great! Adding to Jon's comments - I wonder if the preview field style is distinct enough from editable text fields - not sure if EUI provides a background colour that's clearly distinct from both editable text fields (i.e. white) or the form body colour itself? (Or if there's another styling option that will differentiate from both the form and the editable fields). Also happy to pair!

rhystmills:
Ah, excellent - clear distinction from the background and the editable field, and the toggle button looks great too.

jonathonherbert (approved):
Tested locally, looks _really_ smart. Great work!

PR #362, opened 2023-07-05
Surface 'unpublished changes' status in table

jonathonherbert:
## What does this change?

Surface the 'unpublished changes' status in the rules table, as well as the rule form:
- Adds an icon to both to make the meaning of the symbol clear
- There's a tooltip on hover to explain what the icon is.

Screenshot 2023-07-05 at 12 51 02

One thing to note – at the moment, the icon won't appear in the table until the rules table refreshes. 

We don't refresh the rule table on every update in real-time mode, as it's currently 8MB of JSON, and we don't want to be fetching and parsing that every few seconds. There's a question as to how we keep in sync which probably involves optimistically updating the rule list on update. We could implement that in this PR if we felt that behaviour was a blocker.

## How to test

- Automated tests should pass
- Have a play. Does the new icon appear as expected in the table?

### DIFF
--- a/apps/rule-manager/app/db/DbRuleDraft.scala
+++ b/apps/rule-manager/app/db/DbRuleDraft.scala
@@ -29,7 +29,8 @@ case class DbRuleDraft(
     updatedBy: String,
     revisionId: Int = 0,
     isPublished: Boolean,
-    isArchived: Boolean
+    isArchived: Boolean,
+    hasUnpublishedChanges: Boolean
 ) extends DbRuleCommon {
 
   def toLive(reason: String): DbRuleLive = {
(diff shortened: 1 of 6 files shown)


rhystmills (approved):
Works nicely!

How simple do you think it would be to add a flag per field in the editor in a follow-up PR? Would be nice to know which field has been edited and, as a follow-up to that, revert back to its state in the most recent published rule.

I reckon it would be better to have the 'Live' status and the icon as part of the same fixed-width component as the icon ends up quite far away on tablet & mobile layouts:

image

jonathonherbert:
Thanks @rhystmills, added in 96a2ab6d480b44c0aca9c4d54277fcb2b8fea2be: 

Screenshot 2023-07-13 at 17 04 53

PR #363, opened 2023-07-05
Pass loading state to tag selector, rather than inferring loading via empty list

jonathonherbert:
## What does this change?

Passes loading state to tag selector, rather than inferring loading via empty list, which corrects a bug where when the tool is first loaded without rules or tags, the tag field is populated by a loading indicator (rather than an empty selector.)

## How to test

Run the manager locally with a clean DB. Create a rule. The tags selector should be present, but empty:

Screenshot 2023-07-05 at 13 01 27


samanthagottlieb (approved):
Looks good to me! I've tested locally and it works as expected.

PR #364, opened 2023-07-05
Add client routing

rhystmills:
## What does this change?

This PR adds client routing to the application.

To demonstrate the feature, you can turn on the `Show tags page` feature switch in the top right of the client-side application, which will add a navbar to the header, allowing you to navigate between pages.

image

Any top-level path will now be handed to the client-side React application, unless it is specifically identified in the Play `routes` file (e.g. `/oauthCallback`).

Our other existing routes are now moved to `/api/[path here]`.

I've also added a charming 404 page in the unlikely event that someone types in a non-existent top-level path.

image
 
## How to test

1. Run typerighter locally according to the instructions in the readme
2. Try standard rule creation and publication steps. Do you get any server errors in your network tab? This may indicate that an api path hasn't been updated.
3. Try navigating directly to the url `/tags` via your browser url bar. Does it work?
4. Activate the 'Show tag page' feature switch and try clicking the nav links. Do they take you to expected page?
5. Try navigating directly to the url `/oauthCallback` via your browser. Is it being handled by the backend? (It should be!)
6. Try navigating directly to the url `api/rules` via your browser. Do you see some JSON provided by the backend?
7. Try navigating to a top-level fake path, e.g. `/cheese`. Do you see the 404 page?


apps/rule-manager/conf/routes
```diff
@@ -3,49 +3,54 @@
...
-GET     /oauthCallback              controllers.HomeController.oauthCallback()
-# A temporary endpoint to test permissions
-GET     /editPermissionCheck        controllers.HomeController.editPermissionCheck()
+GET     /                               controllers.HomeController.index(path = "")
+GET     /oauthCallback                  controllers.HomeController.oauthCallback()
 
-+anyhost
-GET     /healthcheck                controllers.HomeController.healthcheck()
++anyhostå
+GET     /api/healthcheck                controllers.HomeController.healthcheck()
```
Fweddi, line 10: The default healthcheck endpoint is /healthcheck (https://github.com/guardian/typerighter/blob/main/cdk/lib/__snapshots__/index.test.ts.snap#L1515) - so unless we remove the `api` prefix here, or adjust the healthcheck endpoint in cdk, I think we'll see a failed deployment (as the instance will be marked as unhealthy).

rhystmills: Ah good point, there's no reason that can't be a reserved endpoint like `oauthCallback` - will update now.

apps/rule-manager/client/src/ts/components/layout/Header.tsx
```diff
@@ -46,9 +54,24 @@ export const Header = () => {
...
-      </HeaderLogo>
+      <NavContainer>
+        <HeaderLogo>
+          <Link to="/">
+            <Logo />
+          </Link>
+        </HeaderLogo>
+        {
+        getFeatureSwitchValue('show-tags-page') ? 
+          <EuiHeaderLinks aria-label="App navigation dark theme example">
```
Fweddi, line 65: Is this aria-label copied from elsewhere?

apps/rule-manager/client/src/ts/components/PageNotFound.tsx
```diff
@@ -0,0 +1,77 @@
...
+import typey from "../../images/typey.gif"
+import typeyArrow from "../../images/typey-arrow.gif"
+import styled from "@emotion/styled"
+
+const TypeyContainer = styled.div`
+    display: flex;
+    flex-direction: horizontal;
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 1.1rem;
+    image-rendering; pixelated
```
Fweddi, line 13: Is this a typo?

apps/rule-manager/client/src/ts/components/PageNotFound.tsx
```diff
@@ -0,0 +1,77 @@
...
+`
+
+const TypeyArrow = styled.img`
+    position: absolute;
+    left: -13px;
+    bottom: 11px;
+    background: none;
+    content: "";
+    transform: scaleY(-1) rotate(90deg);
+}`
```
Fweddi, line 54: This `}` looks a little rogue

jonathonherbert:
@Fweddi's got this review, just chiming in to show my appreciation for the incredible 404 page

apps/rule-manager/conf/routes
```diff
@@ -3,49 +3,54 @@
...
+GET     /api/rules                      controllers.RulesController.list()
 +nocsrf
-POST    /rules                      controllers.RulesController.create()
+POST    /api/rules                      controllers.RulesController.create()
 +nocsrf
-GET    /rules/batch/:ids           controllers.RulesController.getRules(ids: String)
+GET    /rules/batch/:ids                controllers.RulesController.getRules(ids: String)
 +nocsrf
-POST    /rules/batch                controllers.RulesController.batchUpdate()
+POST    /rules/batch                    controllers.RulesController.batchUpdate()
```
Fweddi, line 20: Is there a reason these routes don't have the prefix? They came in recently so this might have slipped through the cracks

Fweddi:
> @Fweddi's got this review, just chiming in to show my appreciation for the incredible 404 page

The PR deserves instant approval for the 404 page alone

apps/rule-manager/conf/routes
Fweddi: (the corresponding fetches would need to be updated too)

Fweddi (approved):
Looks great! I've left a few comments for your discretion, but overall this is a great little piece of work. 

I ❤️ typey!

apps/rule-manager/client/src/ts/components/PageNotFound.tsx
rhystmills: Yes, now removed.

apps/rule-manager/client/src/ts/components/PageNotFound.tsx
rhystmills: The lack of syntax highlighting in those emotion css strings has revealed my inability to type

Fweddi: The only reason I saw these was because IntelliJ highlighted them for me, so maybe there's something you can do with your local setup? 🤷

PR #366, opened 2023-07-06
Add DB instructions to readme

jonathonherbert:
## What does this change?

Adds instructions for connecting to a remote DB to the README.

(I've added a few links and params for convenience, but it's nothing that we don't already encode in CDK!)

## How to test

Useful? Anything to add?

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,3 +85,17 @@ For intellij there is a guide to set up automated linting on save [here](https:/
 ## Automatic Linting
 
 The project contains a pre-commit hook which will automatically run the linter on all staged files. To enable this, run `./script/setup-hooks` from the root of the project.
+
+## Developer how-tos
+
+### Connecting to the rule-manager database in CODE or PROD
+
+Sometimes it's useful to connect to the databases running in AWS to inspect the data locally.
+
+We can use `ssm-scala` to create an SSH tunnel that exposes the remote database on a local port. For example, to connect to the CODE database, we can run:
+
+```bash
+ssm ssh -x -t typerighter-rule-manager,CODE -p composer --rds-tunnel 5000:rule-manager-db,CODE
+```
+
+You should then be able to connect the database on `localhost:5000`. You'll need to use the username and password specified in [AWS parameter store](https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/?region=eu-west-1&tab=Table) at `/${STAGE}/flexible/typerighter-rule-manager/db.default.username` and `db.default.password`.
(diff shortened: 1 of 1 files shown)


samanthagottlieb (approved):
Thanks for these instructions. I've followed them and been able to connect to the CODE database via IntelliJ 🥫 (closest emoji I could find to a database).

Quick question about when to disconnect - should I terminate the connection to the database / close the SSH tunnel once I'm finished doing what I need to do / check in CODE? Or is it ok to keep the connection running in the background?

README.md
```diff
@@ -85,3 +85,17 @@ For intellij there is a guide to set up automated linting on save [here](https:/
...
+
+Sometimes it's useful to connect to the databases running in AWS to inspect the data locally.
+
+We can use `ssm-scala` to create an SSH tunnel that exposes the remote database on a local port. For example, to connect to the CODE database, we can run:
+
+```bash
+ssm ssh -x -t typerighter-rule-manager,CODE -p composer --rds-tunnel 5000:rule-manager-db,CODE
+```
+
+You should then be able to connect the database on `localhost:5000`. You'll need to use the username and password specified in [AWS parameter store](https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/?region=eu-west-1&tab=Table) at `/${STAGE}/flexible/typerighter-rule-manager/db.default.username` and `db.default.password`.
```
jonathonherbert, line 101:
```suggestion
You should then be able to connect the database on `localhost:5000`. You'll need to use the username and password specified in [AWS parameter store](https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/?region=eu-west-1&tab=Table) at `/${STAGE}/flexible/typerighter-rule-manager/db.default.username` and `db.default.password`.

Don't forget to kill the connection once you're done! Here's a handy one-liner `kill $(lsof -ti {PORT_NUMBER})`
```

samanthagottlieb:
> This is good. Databases are now 🥫 .
> 
> One thing that you'll always know as a result of running this command is the local port you've opened. So we can find the PID of the process running on that port (in the example, port 5000), with `lsof -ti :5000`, which'll be the running SSH command, and kill it `kill {PID}` (where `{PID}` is the value returned from the `lsof` command), with this one liner:
> 
> ```
> kill $(lsof -ti 5000)
> ```
> 
> I've added that to the README in [06d9b6e](https://github.com/guardian/typerighter/pull/366/commits/06d9b6e78479e0ea5b824294c723d44f0073e4fd), thanks for the feedback!

Thanks for this!

PR #371, opened 2023-07-10
Layout changes

jonathonherbert:
## What does this change?

A few layout changes. The form content is now scrollable, and the action buttons at the bottom are sticky. The table contents is also scrollable when it exceeds the size of the screen.

The search bar shouldn't scroll with the table, but it's a part of the table component – we've plans to move to another component for rendering our table, so rather than work hard at that, it's probably better to leave it and address when we switch components.

I recommend reviewing with [whitespace changes hidden](https://github.com/guardian/typerighter/pull/371/files?diff=unified&w=1), as there's a bit of shuffling components around and indenting that adds noise to the diff.

layout

## How to test

Have a play. Does the new layout work at different viewport sizes? Any edge cases that'd be good to address?


apps/rule-manager/client/src/ts/components/RulesTable.tsx
```diff
@@ -202,102 +202,108 @@ const RulesTable = () => {
...
-  
+
   const handleRefreshRules = async () => {
     await refreshRules();
     await fetchTags();
   }
 
   return <>
+    <EuiFlexGroup direction="column" gutterSize="none" style={{ height: "100%"}}>
+      <EuiFlexItem grow={0}>
```
samanthagottlieb, line 213: Is `grow={0}` the equivalent of `grow={false}`? I've been using `grow={false}` and wonder if it's preferable to use `grow={0}` for some reason. It does look more intuitive when we also have `grow={1}`, etc. for other components.

apps/rule-manager/client/src/ts/components/layout/Page.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { EuiPageTemplate, EuiProvider } from "@elastic/eui";
-import { Header } from "./Header";
+import {Header, headerHeight} from "./Header";
```
samanthagottlieb, line 3: I noticed when that my IDE did some automatic formatting, which made my PR a bit noisy (e.g. adding spaces around the imports). I thought about adding some prettier config to the project to enforce some consistent formatting, if that would be helpful?

samanthagottlieb (approved):
This is great! I ran it locally and the new scrolling behaviour is very satisfying. Out of interest what are the motivations behind plans to move to another component for rendering the table?

apps/rule-manager/client/src/ts/components/layout/Page.tsx
samanthagottlieb: Will do now 👍

jonathonherbert:
@samanthagottlieb it's something we've discussed ad-hoc but doesn't have a card yet – I'll write it up shortly 👍

PR #378, opened 2023-07-13
Improve linking story

Fweddi:
Suppress Vite logs unless they are of level WARN. This suppresses the logs which link out to localhost ports. As we use dev-nginx, it's misleading to developers to invite them to click on links that open localhost addresses - better to log the dev-gutools domain.

Ideally, we would also suppress the Akka logging around this, and log to the shell when the server is ready, rather than at the start - but I can't find a way to do this.


jonathonherbert (approved):
Tested locally, love this! 👏

PR #379, opened 2023-07-13
Pin SASS version to reduce log noise

Fweddi:
For versions >1.32, SASS spits out deprecation warnings for certain calc() functions. These warnings are from our dependency on ElasticUI, who currently have no intention to fix them - see: https://github.com/elastic/eui/pull/5682#issuecomment-1064378362. 

The best solution until ElasticUI have fixed these warnings is to pin our SASS version.

I've tested locally and the application runs as expected.

### DIFF
--- a/apps/rule-manager/client/package.json
+++ b/apps/rule-manager/client/package.json
@@ -20,7 +20,7 @@
     "@types/react-dom": "~17.0.19",
     "@vitejs/plugin-react-swc": "^3.0.0",
     "jest": "^29.5.0",
-    "sass": "^1.60.0",
+    "sass": "~1.32.13",
     "ts-jest": "^29.1.0",
     "typescript": "^5.1.3",
     "utility-types": "^3.10.0",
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):

apps/rule-manager/client/package.json
```diff
@@ -20,7 +20,7 @@
     "@types/react-dom": "~17.0.19",
     "@vitejs/plugin-react-swc": "^3.0.0",
     "jest": "^29.5.0",
-    "sass": "^1.60.0",
+    "sass": "~1.32.13",
```
Fweddi, line 23: We have tried to suppress the warnings in https://github.com/guardian/typerighter/blob/main/apps/rule-manager/client/vite.config.ts#L29-L33, but this doesn't work because of some decisions by the maintainers of SASS. 

> 1. They [don't want](https://github.com/sass/node-sass/pull/2587#issuecomment-459523943) anyone to ever be able to turn off deprecation warnings, because as soon as those features are removed from the language and go from warnings to errors, they will be [flooded with GitHub issues](https://github.com/sass/libsass/issues/2822#issuecomment-482914373).
> 2. They made one exception, which is if a dependency is the cause, then you can "quiet" just those.
> 3. But then they made it so that their [definition of a "dependency" is](https://sass-lang.com/documentation/js-api/interfaces/StringOptionsWithoutImporter#quietDeps) so strict that it is unlikely anyone actually meets that criteria, so you can't actually turn off the deprecation warnings (their intended outcome).

See https://stackoverflow.com/a/75025363 for more.

PR #382, opened 2023-07-14
Add a method to the checker service to test a single rule

jonathonherbert:
_PR 1/3 to to fulfil the workflow described in https://github.com/guardian/typerighter/discussions/344._

## What does this change?

Adds a method to the checker service to test a single rule, and exposes an endpoint to call it in the controller.

Here's an example of a valid request:

```json
{
    "requestId": "1650f358-4dfe-4f74-aa4b-1d07ae4478ea",
    "rule": {
        "id": "0298e1b6-e28d-4091-a9b4-0c0ff1d3450f",
        "category": {
            "id": "Style guide and names",
            "name": "Style guide and names"
        },
        "description": "MP last elected in 2019: Conservative, South West Surrey",
        "suggestions": [],
        "replacement": {
            "type": "TEXT_SUGGESTION",
            "text": "Jeremy Hunt"
        },
        "regex": "Jeremy Hunt",
        "_type": "com.gu.typerighter.model.RegexRule"
    },
    "documents": [
        {
            "id": "business/2023/jul/13/uk-economy-shrinks-by-01-in-may-despite-coronation-festivities",
            "blocks": [
                {
                    "id": "elem-0",
                    "text": "The chancellor, [Jeremy Hunt](https://www.theguardian.com/politics/jeremy-hunt), said the best way to get growth going again was to ease pressure on households by bringing down inflation. “While an extra bank holiday had an impact on growth in May, high inflation remains a drag anchor on economic growth,” he said.",
                    "from": 0,
                    "to": 413
                }
            ]
        }
    ]
}
```

## How to test

- The new automated tests should pass, and you should be convinced they adequately test the new method
- Running locally or in CODE, run a check with the example request above. It should give some matches as we'd expect.

To succesfully request a check, you'll need to add a pan-domain auth cookie to the request. [Postman](https://www.postman.com/) can be really useful here – this is perhaps most easily done by right-clicking a valid request to checker.typerighter.{stage}.dev-gutools.co.uk, right clicking, and selecting 'copy as curl'. You can then [import the request](https://www.postman.com/postman/workspace/postman-team-collections/folder/1559645-a1d259e3-29ee-4c4c-814c-3a8e2ec71118?ctx=documentation#:~:text=Find%20the%20request%20that%20you,then%20select%20Copy%20as%20cURL.&text=In%20the%20Postman%20app%2C%20click,here%2C%20and%20confirm%20the%20import.) (which will have a valid cookie) in Postman, and alter it for the correct URL and verb (`POST /checkSingle`).

### DIFF
--- a/apps/checker/app/utils/JsonHelpers.scala
+++ b/apps/checker/app/utils/JsonHelpers.scala
@@ -1,0 +1,10 @@
+package utils
+
+import java.net.URL
+import play.api.libs.json.{JsString, Json, Writes}
+
+object JsonHelpers {
+  implicit val urlWrites: Writes[URL] = (o: URL) => JsString(o.toString)
+  def toNDJson[T](serializable: T)(implicit tjs: Writes[T]) =
+    Json.toJson(serializable).toString() + 31.toChar
+}
--- a/apps/checker/app/utils/JsonHelpers.scala
+++ b/apps/checker/app/utils/JsonHelpers.scala
@@ -4,7 +4,9 @@ import java.net.URL
 import play.api.libs.json.{JsString, Json, Writes}
 
 object JsonHelpers {
+  val recordSeparatorChar = 31.toChar
+
   implicit val urlWrites: Writes[URL] = (o: URL) => JsString(o.toString)
   def toNDJson[T](serializable: T)(implicit tjs: Writes[T]) =
-    Json.toJson(serializable).toString() + 31.toChar
+    Json.toJson(serializable).toString() + recordSeparatorChar
 }
--- a/apps/checker/app/AppComponents.scala
+++ b/apps/checker/app/AppComponents.scala
@@ -112,7 +112,7 @@ class AppComponents(
   )(matcherPoolDispatcher, materializer)
 
   val bucketRuleResource = new BucketRuleResource(s3Client, typerighterBucket, stage)
-  val ruleProvisioner = new RuleProvisionerService(
+  val matcherProvisionerService = new MatcherProvisionerService(
     bucketRuleResource,
     matcherPool,
     languageToolFactory,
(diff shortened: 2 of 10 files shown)


apps/checker/app/utils/JsonHelpers.scala
```diff
@@ -0,0 +1,10 @@
+package utils
+
+import java.net.URL
+import play.api.libs.json.{JsString, Json, Writes}
+
+object JsonHelpers {
+  implicit val urlWrites: Writes[URL] = (o: URL) => JsString(o.toString)
+  def toNDJson[T](serializable: T)(implicit tjs: Writes[T]) =
+    Json.toJson(serializable).toString() + 31.toChar
```
Fweddi, line 9: Granted this is old code, but it's difficult to parse. Does 31 represent a \n character or such?

Fweddi (approved):
Tested locally, all LGTM. Have stepped through each commit and code looks sound!

apps/checker/app/utils/JsonHelpers.scala
jonathonherbert: Agreed! It's a record separator char – I've stored it in a variable to avoid a comment in [bf3efd8](https://github.com/guardian/typerighter/pull/382/commits/bf3efd82e624bc25505bea9962d0c53b90ea9dde).

Worth noting that there's another mistake here, in that this is _not_ ndjson, it's [json-seq!](https://en.wikipedia.org/wiki/JSON_streaming#Record_separator-delimited_JSON) There are a few flavours of streaming json, confusingly. Rather than rename in this PR, I'll address in the next PR or so (there are a few changes to this file coming that add extra streaming utils, and the refactor touches a few files which will all need rebasing.)

PR #385, opened 2023-07-14
Test a rule from the rule manager service with CAPI content

jonathonherbert:
_NB: Don't be afraid of the large diff – most of it is a chunky CAPI fixture! 😀_

## What does this change?

Adds an endpoint to test a rule from the rule manager service with CAPI content. We can supply a query string, tags and sections to the query.

The route is GET `/api/rules/:id/test-capi`. An example body might look like:

```
{"queryStr": "Jeremy Hunt", "tags": ["politics/jeremy-hunt"], "sections": []}
```

An example response looks like:

```
{
  "currentPage": 1,
  "maxPages": 100,
  "result": {
    "matches": [
      {
        "rule": {
          "id": "0298e1b6-e28d-4091-a9b4-0c0ff1d3450f",
          "category": {
            "id": "Style guide and names",
            "name": "Style guide and names"
          },
          "description": "MP last elected in 2019: Conservative, South West Surrey",
          "suggestions": [],
          "replacement": {
            "type": "TEXT_SUGGESTION",
            "text": "Jeremy Hunt"
          },
          "regex": "Jeremy Hunt",
          "_type": "com.gu.typerighter.model.RegexRule"
        },
        "fromPos": 3906,
        "toPos": 3917,
        "precedingText": "The chancellor, ",
        "subsequentText": ", said the best way to ease the pressure on households was to get inflation falling further. “We are",
        "matchedText": "Jeremy Hunt",
        "message": "MP last elected in 2019: Conservative, South West Surrey",
        "shortMessage": "MP last elected in 2019: Conservative, South West Surrey",
        "suggestions": [],
        "replacement": {
          "type": "TEXT_SUGGESTION",
          "text": "Jeremy Hunt"
        },
        "markAsCorrect": true,
        "matchContext": "The chancellor, [[Jeremy Hunt]], said the best way to ease the pressure on households was to get inflation falling further. “We are"
      }
    ],
    "percentageRequestComplete": 10
  }
}
```

## Dev notes

We fulfil the workflow described in the discussion, which looks like:

```mermaid
sequenceDiagram
    participant RMC as Rule management client
    participant RMS as Rule management server 
    participant CS as Checker server
    participant CAPI
    RMC ->> RMS: [GET] checkerRule, capiQuery?, matchCount
    Note over RMC, RMS: Query optional, defaults to latest content
        loop for each page of CAPI, until matchCount matches found, we run out of pages, or some limit of pages reached
        RMS -->> CAPI: [GET] capiQuery
        CAPI -->> RMS: content[]
        Note over RMS: Convert HTML content to block[]
        loop for each page of documents
            RMS ->> CS: [GET] checkerRule, document[]
            CS ->> RMS: [Chunked json-seq] match[]
        end
    end
    Note over RMS: Stream matches back to client, filtering documents with no matches and reporting progress
    RMS ->> RMC: [Chunked json-seq] progress, match[]
```

We're using Akka streams to do this in `RuleTesting.scala`. It's the most declarative way I can think to manage that workflow in a way that's robust. Given that we're already streaming the response from the checker, we can combine streams in a way that's quite elegant.

I've tried to comment the stream logic to make it clear which part is doing what, as the DSL for akka-streams is quite specific, but super happy to pair to talk things through, and if we can improve the comments to make it clearer, _don't hold back._

## How to test

- The automated tests should run. I've mocked a CAPI client in `RuleTestingSpec` with a JSON response pulled from a live query for ... verisimilitude.
- Have a play with some queries! Running locally or in CODE, pull some rules from the Google sheet, take note of the ID of a specific rule, and make a query against the endpoint (don't forget to get a valid auth cookie, per the instructions in [this PR](https://github.com/guardian/typerighter/pull/382)) with the URL and body in the description. You should be able to see results come in as in the response above.


apps/rule-manager/app/service/RuleTesting.scala
```diff
@@ -1,20 +1,113 @@
...
+      matchCount: Int = 10,
+      maxPageCount: Int = 100
+  )(implicit ec: ExecutionContext): Source[PaginatedCheckRuleResult, NotUsed] = {
+    class PaginatedContentQuery(pageLimit: Int) {
+      var currentPage = 1
+      def hasNext = currentPage <= pageLimit
+      def nextPage(): Future[(Int, List[Document])] = {
+        log.info(s"Fetching content page $currentPage/$maxPageCount")
+        val pageToFetch = currentPage
+        currentPage = currentPage + 1
```
Fweddi, line 61: 👽 I spot mutation!

PR #386, opened 2023-07-14
Log play domains on server start

Fweddi:
## What does this change?

I spent a while trying to work this out. It's not in the Play docs, but it is in @andrew-nowak's head. Let that be a lesson, folks.

This uses a simple little play hook to spit out the local nginx domain when each app is ready. We could have opted to open the domain too, as per https://github.com/guardian/birthdays/blob/eea8c84f9938065fec33b9dad03af7642fb37705/project/PlayVite.scala#L60, but we don't necessarily want to open both apps, and sometimes it can be annoying to have a new window opened if you're already working on it.

Hopefully this makes the dev start experience just that little bit quicker.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,12 +69,13 @@ val commonLib = (project in file(s"$appsFolder/common-lib"))
     )
   )
 
-def playProject(projectName: String, devHttpPorts: Map[String, String]) =
+def playProject(label: String, projectName: String, domainPrefix: String, devHttpPorts: Map[String, String]) =
   Project(projectName, file(s"$appsFolder/$projectName"))
     .dependsOn(commonLib)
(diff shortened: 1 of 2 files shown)


jonathonherbert:
This is amazing! Just to double check, did you see https://www.playframework.com/documentation/2.8.x/sbtCookbook? Might be useful for folks coming across this PR, but perhaps you spotted it and it doesn't contain what you needed.

Fweddi:
> This is amazing! Just to double check, did you see [playframework.com/documentation/2.8.x/sbtCookbook](https://www.playframework.com/documentation/2.8.x/sbtCookbook)? Might be useful for folks coming across this PR, but perhaps you spotted it and it doesn't contain what you needed.

I did not - but I think Andrew mentioned it. Thanks!

PR #389, opened 2023-07-20
Check a rule from the rule manager

jonathonherbert:
## What does this change?

Adds an endpoint to check a rule manager rule from the rule manager against the specified copy.

The route is `/api/rules/:id/test-block`, and the document is specified in the body – here's an example:

```json
{
  "id": "id",
  "blocks": [
    {
      "id": "elem-0",
      "text": "The chancellor, [Jeremy Hunt](https://www.theguardian.com/politics/jeremy-hunt), said the best way to get growth going again was to ease pressure on households by bringing down inflation. “While an extra bank holiday had an impact on growth in May, high inflation remains a drag anchor on economic growth,” he said.",
      "from": 0,
      "to": 413
    }
  ]
}
```

## Dev notes

- We stream matches directly from the checker. 
- We're using HMAC auth to make the request to the checker from the rule-manager.

## How to test

- The automated tests should pass, and you should be convinced they cover the testing method. We simulate a server in the tests to test the streaming behaviour, which makes them quite comprehensive.
- Check content yourself locally! A GET request to `https://manager.typerighter.local.dev-gutools.co.uk/api/rules/:id/test-block` (be sure to add a valid cookie in the request) with the payload above should give a `200` response.


apps/rule-manager/app/service/RuleTesting.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class RuleTesting(
+    ws: WSClient,
+    hmacClient: HMACClient,
+    checkerUrl: String
+) extends Loggable {
+
+  /** Test a rule against the given list of documents. Return a list of matches from the checker
+    * service endpoint as streamed NDJSON.
```
jonathonherbert:
```suggestion
    * service endpoint as streamed json-seq.
```

apps/rule-manager/test/service/RuleTestingSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class RuleTestingSpec extends AnyFlatSpec with Matchers with IdiomaticMockito {
+
+  val as: ActorSystem = ActorSystem()
+  implicit val materializer: Materializer = Materializer(as)
+
+  /** Mock responses from our CAPI client and checker service when parsing responses.
```
jonathonherbert:
```suggestion
  /** Mock responses from the checker service.
```

apps/checker/conf/application.conf
```diff
@@ -11,6 +11,7 @@ play.assets {
 typerighter {
   sheetRange="A:G"
   defaultAwsProfile = "composer"
+  checkerServiceUrl = "typerighter.local.dev-gutools.co.uk"
```
Fweddi, line 14: I think the url should include the prefix here - **checker**.typerighter.local.dev-gutools.co.uk.

But I can't see where this config value is used, and I'm not sure if this is how we should be using config? 

This config describes the checker app's own features, rather than configuring the checker app to use a particular property. 

It makes sense when we use it in the manager's config, but I'm not sure it makes sense here.

apps/rule-manager/app/controllers/RulesController.scala
```diff
@@ -190,4 +194,16 @@ class RulesController(
         }
     }
   }
+
+  def testWithBlock(id: Int) = APIAuthAction[JsValue](parse.json).async { implicit request =>
+    request.body.validate[Document].asEither match {
+      case Right(document) =>
+        val rule = DbRuleDraft.find(id).get
```
Fweddi, line 201: At one point in the commit history we handled the rule's optionality, but it seems we've lost this?

apps/rule-manager/app/service/RuleTesting.scala
```diff
@@ -0,0 +1,59 @@
...
+    ws: WSClient,
+    hmacClient: HMACClient,
+    checkerUrl: String
+) extends Loggable {
+
+  /** Test a rule against the given list of documents. Return a list of matches from the checker
+    * service endpoint as a stream of json-seq records.
+    */
+  def testRule(rule: DbRuleDraft, documents: List[Document])(implicit ec: ExecutionContext) = {
+    val liveRule = rule.toLive("placeholder")
```
Fweddi, line 22: Heh, a little strange but I understand why!

Fweddi: I guess we could add a method `draftDbRuleToCheckerRule`, but it's not a blocker (a new method might not be as DRY).

Fweddi (commented):
Tested and works as expected! Nice one.

Have left a few comments - holding off on approval until CODE is green.

Fweddi (approved):
Tested locally and on CODE and it works as expected! Great stuff.

Have left a few non-blocking comments.

apps/rule-manager/app/service/RuleTesting.scala
jonathonherbert: Yeah, it's a bit of a pain, this. Be a nice refactor down the line.

apps/rule-manager/app/controllers/RulesController.scala
jonathonherbert: Thanks for spotting! Added in 4cde00e17460cfc26e3ad56825341ab42bf25621.

apps/checker/conf/application.conf
jonathonherbert: Good catch. This was likely left behind when I thought that the entire request URL was signed by HMAC. Turns out, it's just the URL's path, so this isn't necessary – removed in 0b64a29ecf9ce4c9189225214d5cf0bf02197c5c.

PR #390, opened 2023-07-20
Remove setting up node version from script

samanthagottlieb:
## What does this change?

This removes the steps that try to set up the correct node version from the `check-node-version` script. These changes were added [here](https://github.com/guardian/typerighter/pull/384). However they were preventing me from running the `start-manager` locally, with the script exiting at [this line](https://github.com/guardian/typerighter/pull/384/files#diff-118f9a380506b6024afd5a8be1c0133756597d95cae682bcc995b1c1de0fc0cdR4):
```
[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
```
@Fweddi and I looked at this together and didn't find the exact cause of the issue. Rather than adding more code to handle whether someone has fnm, nvm, asdf etc. and where it might be installed (e.g. homebrew installs it somewhere specific), we thought it would be better to remove this code and simply notify someone they're using the wrong node version, allowing them to change versions manually themselves.

Now if someone tries to run the script locally and they have the wrong version, the script will simply print an error message letting them know they're using wrong version of node and will then exit. This matches behaviour if you're using the wrong java version. This PR also changes the exit code if you have the wrong java version to reflect the operation not being permitted.

## How to test

Change node versions and try running the `start-manager` or `start` script. You should see an error message explaining you're using the wrong node version and the script should exit.

### DIFF
--- a/script/lib/check-java-version
+++ b/script/lib/check-java-version
@@ -13,7 +13,7 @@ checkJavaVersion() {
 
   if [ "$runningJavaVersion" != "$requiredJavaVersion" ]; then
     echo -e "Using wrong version of Java. Required ${requiredJavaVersion}. Running ${runningJavaVersion}."
-    exit 0
+    exit 1
   fi
 }
 
--- a/script/lib/check-node-version
+++ b/script/lib/check-node-version
@@ -1,52 +1,7 @@
 #!/bin/bash -e
 
-export NVM_DIR="$HOME/.nvm"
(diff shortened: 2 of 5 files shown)


Fweddi (approved):
Tested locally, LGTM! Nice spot with the exit codes.

While we're here, I've also added a commit to change the `.` command to `source`. `.` is the alias for `source`, but more obscure and difficult to look up - I had some feedback to change these in #384, but missed out these references.

(these aren't related to your changes, I'm just piggybacking!)

PR #391, opened 2023-07-21
Add a diff of pending changes to the rule publication modal

rhystmills:
## What does this change?

It's currently not clear what changes are being published when publishing a new version of a live rule.

This PR adds a diff to the publication modal to show what fields have been updated, plus tests for the functions added.

image

## How to test

1. Run the rule manager locally according to the instructions in the readme.
2. Change some fields in a Live rule (If you have no Live rules, publish a Draft rule. If you have no draft rules, pull rules from the Google Sheet by activating the feature switch in the top right dropdown menu).
3. Do you receive a diff of any field you change? Do any fields show up in the diff when they shouldn't?

### DIFF
--- a/apps/rule-manager/client/src/ts/components/Diff.tsx
+++ b/apps/rule-manager/client/src/ts/components/Diff.tsx
@@ -1,0 +1,165 @@
+import { EuiBadge, EuiFlexGroup, EuiFlexItem, EuiHorizontalRule, EuiSpacer, EuiText, useEuiTextDiff } from "@elastic/eui";
+import { BaseRule, RuleData } from "./hooks/useRule";
+import { useTags } from "./hooks/useTags";
+import { css } from "@emotion/react";
+import styled from "@emotion/styled";
+
+type DivergentFields = {
+    fieldName: string;
+    live: FieldValue;
+    draft: FieldValue;
+}
+
+type FieldValue = string | number | boolean | number[] | string[] | undefined
+
+type FieldObject = {
+    fieldName: string,
+    value: FieldValue
+}
+
+const ComparisonPanel = styled.div`
+    background-color: #F1F4FA;
+    height: 100%;
+    padding: 0.5rem;
+`
+
+const ComparisonPanelHeader = styled.div`
+    padding: 0.25rem 0;
+    color: #444;
+`;
+
+const diffWrapper = css`
+    border: 1px solid #ddd;
+    padding: 1rem;
+    border-radius: 0.5rem;
+`;
+
+const textDiffFields = ["description", "pattern", "replacement"];
+const comparisonDiffFields = ["ruleType", "category", "tags"];
+
+const doValuesMatch = (draftValue?: FieldValue, liveValue?: FieldValue): boolean => {
+    // The value could be an array or a primitive
+    if (Array.isArray(draftValue) && Array.isArray(liveValue)){
+        // This is a good enough comparison for our arrays of primitives
+        return JSON.stringify(draftValue) === JSON.stringify(liveValue)
+    }
+    return draftValue === liveValue
+}
+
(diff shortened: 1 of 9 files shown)


apps/rule-manager/client/src/ts/components/Diff.tsx
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return field
+    })
+}
+
+export const Diff = ({rule}: {rule: RuleData | undefined}) => {
+    const {tags} = useTags();
+    
+    const diffedFields = rule ? transformToHumanReadableValues(findFieldsWithDiffs(rule), tags) : [];
+    const textDiffs = diffedFields.filter(diffedField => textDiffFields.includes(diffedField.fieldName))
```
jonathonherbert, line 115: Non-blocking – I'd be tempted to move the logic that cared about what sort of diff to provide into a `<FieldDiff` object, which then decided which sort of diff to render. It'd make this component simpler – you could even pass the `DivergentField` straight to the component – which'd then decide which sort of diff to render, and keep all of the diff-related logic together.

apps/rule-manager/client/src/ts/utils/camelCaseToTitleCase.ts
```diff
@@ -0,0 +1,5 @@
+export const camelCaseToTitleCase = (str: string) => {
+    // Add space between strings
+    const result = str.replace(/([A-Z])/g,' $1');
```
jonathonherbert, line 3: Ingenious! But could we replace with lodash's `startCase`?

jonathonherbert (approved):
This is proper nice. We should probably have it for Composer! A few non-blocking comments, but I think we could also ship as-is.

One general comment about the UI – I'm not convinced we need so much detail! On my screen, we ended up with a scrollbar as the UI currently stands, but stripping the extra heading, border and background gave us a bit more space – idk what you think?

|1|2|
|-|-|
|Screenshot 2023-07-24 at 09 18 43|Screenshot 2023-07-24 at 09 18 13|

Appreciate the 'what's changed' heading gives some context. I feel like we should be able to get that in the table headings if it's not already clear – e.g 'before republish', 'after republish', etc., but these aren't strongly held opinions, if you feel differently let's ship and iterate.

apps/rule-manager/client/src/ts/components/Diff.tsx
```diff
@@ -0,0 +1,163 @@
...
+    padding: 1rem;
+    border-radius: 0.5rem;
+`;
+
+const textDiffFields = ["description", "pattern", "replacement"];
+const comparisonDiffFields = ["ruleType", "category", "tags"];
+
+export const doValuesMatch = (draftValue?: FieldValue, liveValue?: FieldValue): boolean => {
+    // The value could be an array or a primitive
+    if (Array.isArray(draftValue) && Array.isArray(liveValue)){
```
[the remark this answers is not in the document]
rhystmills: Delightful, now implemented.

apps/rule-manager/client/src/ts/utils/camelCaseToTitleCase.ts
rhystmills: Smashing, let's see how they did it:
```
const startCase = (string) => (
  words(`${string}`.replace(/['\u2019]/g, '')).reduce((result, word, index) => (
    result + (index ? ' ' : '') + upperFirst(word)
  ), '')
)
```

jonathonherbert: Crikey 😀

apps/rule-manager/client/src/ts/components/Diff.tsx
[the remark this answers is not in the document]
rhystmills: Your later comment made me curious - Lodash also has an `isEqual` function that will do this. Let's save some chars

PR #392, opened 2023-07-24
Speed up the build a bit

jonathonherbert:
## What does this change?

A few things to shorten our build times, which are routinely ~10 minutes at present:
- Test and scalafmt within a single SBT session, rather than multiple sessions.
- Caching for npm and sbt.

## How to test

Check the actions tab. This branch should run a bit quicker:

Screenshot 2023-07-24 at 18 29 58

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,12 +19,15 @@ jobs:
         with:
           java-version: "8"
           distribution: "corretto"
+          cache: "sbt"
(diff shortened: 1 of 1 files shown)


samanthagottlieb (approved):
This is great! So the cache will be stored and reused across subsequent workflow runs as long as the package-lock.json is unchanged?

jonathonherbert:
Exactly, @samanthagottlieb 👍

PR #395, opened 2023-07-25
Pull dictionary xml file from S3 into the Rule Manager

rhystmills:
## What does this change?

This PR gives the Rule Manager access to a dictionary XML file from s3, and parses it to a list of unique words.

More specifically, it:
-  Enables the Rule Manager to fetch the lemmatised dictionary XML file from S3 when a user hits the `refreshDictionaryRules` endpoint. The list is parsed and converted to a list of unique words - for now this list is only returned to the user - in a follow-up PR we will convert the words to rules in our database.
- Adds tests for methods used in the above process.
- Replaces the `setup-hooks` script with a generic `setup` script that runs our `dev-nginx` setup and pulls the dictionary file for local use.
- Modifies our Docker yaml and the localstack `init` script to make the dictionary XML file available in our local S3 instance.

## How to test

Run the rule manager locally.

Send a POST request to [https://manager.typerighter.local.dev-gutools.co.uk/api/refreshDictionary](https://manager.typerighter.local.dev-gutools.co.uk/) - I use [Postman](https://web.postman.co/) to do this, using auth cookies from the network tab https://manager.typerighter.local.dev-gutools.co.uk/ in the browser (I think this just needs to be `gutoolsAuth-assym` but I tend to include `PLAY_SESSION` too).

This should return a list of unique words.

1. Are the words unique?
2. What happens if you do this before running the setup script? (you can delete the dictionary from `~/.gu/typerighter/` if you've already run it)
3. Does it work if you deploy to CODE?


jonathonherbert (approved):
Code looks great! Lovely to see tests. Tested locally, works as expected – I've made a couple of changes to init-aws.sh to get the paths working that I hope makes sense in 8e06f65866f033f9220d5a6366f87336f872aedb.

rhystmills:
> Code looks great! Lovely to see tests. Tested locally, works as expected – I've made a couple of changes to init-aws.sh to get the paths working that I hope makes sense in [8e06f65](https://github.com/guardian/typerighter/commit/8e06f65866f033f9220d5a6366f87336f872aedb).

I think line 9 was sort of right (this one):
```
awslocal s3 cp /etc/gu/typerighter/typerighter-dictionary.xml s3://typerighter-app-local/local/dictionary/typerighter-dictionary.xml
```
... it's copying the file from a location that Docker copies the file to in the localstack container (`etc/.gu...`) as specified in the `volumes` bit of the `docker-compose.yml` - but I've realised the convention is `/etc/gu` rather than `/etc/.gu` so I'll update that.

I should have remove line 8 because it wasn't doing anything - the localstack container was failing to access composer which led me down the Docker `volumes` path.

rhystmills:
(tested that in CODE and locally - working nicely so merged now)

jonathonherbert:
Ace, thanks @rhystmills! I've a memory that adding config to the user dir `~` is more common now (I can see composer, atom workshop and workflow in there), but unsure if others can confirm, it's totally fine your way too 👍

PR #397, opened 2023-07-27
Make FormError keys machine-readable rather than user facing

Fweddi:
## What does this change?

At the moment, we have FormError objects we send back to the client on publication validation failures. Both key and message are human readable.

This PR treats the key as an error id (e. field-missing). Only the message should be human readable. This applies whenever we’re passing FormError events back to the client.

## How to test

Have a play around with the form on the client and see if the form error IDs make sense. 

Tests should pass.

We shouldn't lose necessary error context in this change.

### DIFF
--- a/apps/rule-manager/app/service/RuleManager.scala
+++ b/apps/rule-manager/app/service/RuleManager.scala
@@ -80,6 +80,13 @@ object RuleManager extends Loggable {
     }
   }
 
+  def camelToKebab: String => String = _.foldLeft("") {
+    case (acc, chr) if chr.isUpper => acc :+ '-' :+ chr.toLower
+    case (acc, chr)                => acc :+ chr
+  }
+  def prefixFormError(e: FormError): FormError =
+    FormError(s"invalid-${camelToKebab(e.key)}", e.message)
+
   def liveDbRuleToCheckerRule(rule: DbRuleLive): Either[Seq[FormError], CheckerRule] = {
     rule match {
       case r: DbRuleLive if r.ruleType == RuleType.regex =>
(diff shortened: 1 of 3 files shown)


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -80,6 +80,13 @@ object RuleManager extends Loggable {
     }
   }
 
+  def camelToKebab: String => String = _.foldLeft("") {
```
Fweddi, line 83: This is so we don't have mixed casing like `invalid-externalId`. This isn't an exhaustive translation to kebab-case (e.g. multiple capitals would cause issues), but it does deal with regular camelCase.

rhystmills (commented):
Looking at the source of `FormError` (`play/api/data/FormError.scala`):
```
A form error.
Params:
key – The error key (should be associated with a field using the same key).
messages – The form message (often a simple message key needing to be translated), if more than one message is passed the last one will be used.
args – Arguments used to format the message.
```
.. it looks like the intention is to use this error type for forms and their specific fields - so we're perhaps using it for purposes beyond its intended use across the project. I'd be keen to stick with the field name where we can to match this intended use (e.g. for `pattern`,`description`, `ruleType` - which would e.g. make it easy to provide inline error hints against those fields in the future.

In the other cases your approach looks like an improvement on what we've got currently.

Fweddi:
> 

I suppose it depends on how tightly we think the field should associated with the key. I'd argue in this PR there is still a strong association between the key and the field name?

rhystmills:
I think the idea is that - if the error key and the field name exactly match, you can used the error key from the backend as an identifier for the field in the frontend, then match them up in the frontend to show a field's error directly against the relevant field. However, we're not doing that, and the error messages provided were not very useful when I tried this, plus we have our own inline validation in the client anyway. If we did want to do this, it would just require a minimal transformation from this PR (i.e. remove the "`invalid-`",

rhystmills (approved):
Definitely an improvement on the messy error structure at the moment.

Tested locally and all seems happy 👍

PR #400, opened 2023-08-03
Create live and draft rules for dictionary words on ingest

rhystmills:
## What does this change?

This PR converts dictionary words to live and draft rules on ingest of the dictionary file.

It also provides an interface for listing dictionary rules in the rule manager, this has been designed to work well with a debounced 'search-as-you-type' search similar to the one in the existing rules table:

`/api/dictionary` - > List all rules, showing the first 1000
`/api/dictionary?page=1` -> List all rules, showing the second 1000 (pages are zero-indexed)
`/api/dictionary?word=bel` -> show words starting with 'bel', ordered by word similarity (based on shared trigrams between search term and result word)
`/api/dictionary?word=a&page=2` -> show the third batch of 1000 words starting with 'a'

image

Dictionary rules are not included in the standard `list` endpoint so that they don't clutter the frontend rules table.

The endpoints are paginated for performance reasons - serialising to JSON all the rows for e.g. the search 'a' would be noticeably slow.

In follow-up PRs:
- Dictionary rules will be included in the published rules artefact and ingested by the Checker service, where they will be used in a spellchecker Matcher Pool (they're not included in this PR because they're not 'active' live rules and the there's some additional work required to include them in the artefact)
- We will introduce a frontend interface for listing, unpublishing and editing tags of dictionary rules

## How to test

1. Make sure you've run the 'setup' script to ensure the dictionary XML file is available locally.
2. Run the rule manager locally according to the instructions in the readme
3. Send a request to https://manager.typerighter.local.dev-gutools.co.uk/api/refreshDictionary in order to add the dictionary rules to your local database
4. Make some requests to `/api/dictionary`, e.g. `/api/dictionary?word=bel`. Do they behave as expected?


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -349,4 +355,51 @@ object RuleManager extends Loggable {
       case e: Exception => Left(e)
     }
   }
+
+  def destructivelyPublishDictionaryRules(
+      words: List[String],
+      bucketRuleResource: BucketRuleResource
+  ) = {
+    // Destroy existing draft dictionary rules
+    val ids = DbRuleDraft.findAllDictionaryRules().map(rule => rule.id)
```
jonathonherbert, line 364: I think it's worth writing a method for doing this; we're loading 250k rules into memory here, and then making 250k DB calls, which is ... lots.

We could also make the relation between the tag and the rule ON CASCADE DELETE to make this unnecessary – I think that's reasonable for a join table, where removing the rule will always entail removing the join (and removing rules is very rare!)

apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -203,6 +205,78 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
+            .on(rd.id, rt.ruleId)
+            .where
+            .eq(rd.ruleType, "dictionary")
+            .and
+            .like(rd.pattern, s"$word%")
+            .groupBy(dbColumnsToFind, rl.externalId, rl.revisionId)
+            .orderBy(sqls"sml DESC")
+            .limit(1000)
+            .offset(page * 1000)
+        }.map(DbRuleDraft.fromRow)
```
jonathonherbert, line 247: Non-blocking: suspect there are a few ways of drying this up! For example, we could store the result of the initial `select` call for each case, or store the arguments for the function as a list and conditionally add the `wordSimilarity` column to spread into the call.

Happy for us to ship and refine, esp. if we think our approach to the rule list may change.

apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -350,37 +428,59 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
-        Symbol("createdBy") -> entity.createdBy,
-        Symbol("createdAt") -> entity.createdAt,
-        Symbol("updatedBy") -> entity.updatedBy,
-        Symbol("updatedAt") -> entity.updatedAt,
-        Symbol("isArchived") -> entity.isArchived,
-        Symbol("ruleOrder") -> entity.ruleOrder
-      )
+      hasNoExternalId match {
+        case true =>
+          Seq(
```
jonathonherbert, line 437: Could we have a common `Seq` here, and conditionally append the extra field, saving us the repetition?

jonathonherbert (commented):
Tested locally, and it works well! A few comments / questions.

Generally, my feeling is that we should aim for a single way of querying our rules, and if it's clear that we're not going to be able to store everything in memory (which I think is where we're at with the dictionary work?), we should alter our approach to server-side search and pagination, as you do here – this PR is a neat stepping stone if we do make that decision.

apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -350,37 +428,59 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
-        Symbol("forceRedRule") -> entity.forceRedRule,
-        Symbol("advisoryRule") -> entity.advisoryRule,
-        Symbol("createdBy") -> entity.createdBy,
-        Symbol("createdAt") -> entity.createdAt,
-        Symbol("updatedBy") -> entity.updatedBy,
-        Symbol("updatedAt") -> entity.updatedAt,
-        Symbol("isArchived") -> entity.isArchived,
-        Symbol("ruleOrder") -> entity.ruleOrder
-      )
+      hasNoExternalId match {
```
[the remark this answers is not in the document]
rhystmills: I didn't figure out exactly where it happens but it ends up with a unique external id in the DB so I was confident it was happening somewhere. Something like `536b5404-2a5f-4bb7-b1e5-a48e94212548`.

apps/rule-manager/app/service/RuleManager.scala
rhystmills: Sounds wise, let's do that. To clarify, do you mean between the rule and the rule-tag?

apps/rule-manager/app/db/DbRuleDraft.scala
rhystmills: I seem to remember I tried to mutate and found that `Seq` has a fixed length, but it looks like I could append to create a new Seq in that case, would be nicer.

rhystmills:
> Tested locally, and it works well! A few comments / questions.
> 
> Generally, my feeling is that we should aim for a single way of querying our rules, and if it's clear that we're not going to be able to store everything in memory (which I think is where we're at with the dictionary work?), we should alter our approach to server-side search and pagination, as you do here – this PR is a neat stepping stone if we do make that decision.

Yeah, we won't be able to store everything in memory I think. The client side search has been pleasan, but I can see the argument for consistency. The `LIKE 'suffix%'` SELECT I'm using for dictionary rules might not be as suitable for the other rules, so we might want to rethink that if we're doing server-side search for standard rules too.

apps/rule-manager/app/service/RuleManager.scala
jonathonherbert: Nice, yep, ON CASCADE DELETE for the rule FK in a rule tag.

apps/rule-manager/app/db/DbRuleDraft.scala
rhystmills: (Now done)

apps/rule-manager/app/db/DbRuleDraft.scala
rhystmills: Have reduced a bit of repetition with the arg list but not had much luck with the select repetition due to the nuances of the DSL. Quick pair might be handy!

apps/rule-manager/app/service/RuleManager.scala
jonathonherbert: That'd be fair, could be a follow up!

jonathonherbert (approved):

jonathonherbert (approved):
Looks great! There was trouble with the build, so I've added [3203107](https://github.com/guardian/typerighter/pull/400/commits/32031071851efe67337b90f231637ffc14367050) to work around an awkward problem ScalikeJDBC causes when assigning query builders to vals – see the link in the extended commit message for a description.

jonathonherbert:
Oh – and let's make sure to either backport the `fetchSize` config from the following PR (https://github.com/guardian/typerighter/commit/154e0a97d245a8653eb0488de501b62f4543795d I think), or merge the two PRs together, to avoid OoMs!

rhystmills:
Wise, I've add them here too

PR #403, opened 2023-08-08
A single, lazy-loaded list for rules

jonathonherbert:
## What does this change?

A PoC for a virtualised, lazy-loaded list of rules, as loading a full dictionary into memory runs into network and performance constraints.

## To do

- [x] Spike [react-window](https://github.com/bvaughn/react-window)
- [x] A rough first go at a styled table and search layout
- [x] Multi-selection
- [ ] Select all (limited to n rows)
- [ ] Add full-row index and search endpoint (could be a separate PR)
- [ ] Connect search with lazy-loaded table rows

## How to test

Still to come.


jonathonherbert:
Closed in favour of #426.

PR #404, opened 2023-08-10
Put form error message in right place

jonathonherbert:
## What does this change?

The form error message is in a weird place. This PR fixes.

|Before|After|
|-|-|
|image|Screenshot 2023-08-10 at 10 48 52|

## How to test

Create a rule without a pattern. Does the error message appear in the right place?


rhystmills (approved):
Delightful, tested locally and solved the problem 👍

PR #406, opened 2023-08-11
Validate patterns for regex and xml rules

jonathonherbert:
## What does this change?

Adds validation for regex and LTXML rules. It's surfaced under the field. The validation errors themselves can be rather obscure, but I've tried to make it clear that they're related to the syntax of the pattern.

It all happens server-side.

Also links the validator errors for categories.

|regex|xml|
|-|-|
|Screenshot 2023-08-11 at 13 59 50|Screenshot 2023-08-11 at 13 59 59|

## How to test

Create a rule, and then enter an invalid regex (for example, an unterminated pair of parentheses), or invalid xml (for example, an unterminated tag, or no tag at all!) for regex and languagetool rules respectively. You should get the appropriate error messages

### DIFF
--- a/apps/rule-manager/client/src/ts/components/helpers/errors.ts
+++ b/apps/rule-manager/client/src/ts/components/helpers/errors.ts
@@ -1,6 +1,21 @@
+import { capitalize } from 'lodash';
 import { FormError } from '../RuleForm';
 
-export const hasErrorsForField = (
+const defaultErrors: FormError[] = [];
+
+/**
+ * Given a list of validation errors and a field name, return props that can be
+ * spread in
+ */
+export const getErrorPropsForField = (
 	fieldName: string,
 	formErrors: FormError[] | undefined,
-) => !!formErrors?.some(({ key }) => key === fieldName);
+) => {
+	const fieldErrors =
+		formErrors?.filter(({ key }) => key === fieldName) || defaultErrors;
+
+	return {
+		isInvalid: !!fieldErrors?.length,
+		error: fieldErrors?.map(({ message }) => capitalize(message)),
+	};
+};
--- a/apps/rule-manager/client/src/ts/components/helpers/errors.ts
+++ b/apps/rule-manager/client/src/ts/components/helpers/errors.ts
@@ -1,0 +1,6 @@
+import { FormError } from '../RuleForm';
+
+export const hasErrorsForField = (
+	fieldName: string,
+	formErrors: FormError[] | undefined,
+) => !!formErrors?.some(({ key }) => key === fieldName);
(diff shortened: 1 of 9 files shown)


rhystmills:
I think the `CategorySelector` in the `RuleFormBatchEdit` component also needs access to validationErrors.

jonathonherbert:
Thanks for spotting! Ooh, batch editing publication validation. I won't address that in this PR (it's not addressed at all IIUC), but it's a good reminder that the validation story for that will need to be different. ab6e7143687b35e6d5560e9c41b54b36de9a058c fixes.

apps/rule-manager/client/src/ts/components/helpers/errors.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Given a list of validation errors and a field name, return props that can be
+ * spread in
+ */
+export const getErrorPropsForField = (
+	fieldName: string,
+	formErrors: FormError[] | undefined,
+) => {
+	const fieldErrors =
+		formErrors?.filter(({ key }) => key === fieldName) || defaultErrors;
```
samanthagottlieb, line 15: For some reason IntelliJ underlines this for me - with the error `Argument types do not match parameters`

It's happy with this however:
```
const fieldErrors = formErrors
		? formErrors.filter(({ key }) => key === fieldName)
		: defaultErrors;
```

Also VSCode has no problems with how you've done it, so I imagine it's just an IntelliJ issue!

apps/rule-manager/test/db/RuleManagerSpec.scala
```diff
@@ -88,6 +88,7 @@ class RuleManagerSpec extends FixtureAnyFlatSpec with Matchers with AutoRollback
       maybeCheckerRule shouldBe Left(
         List(
           FormError("invalid-pattern", List("error.required"), List()),
+          FormError("invalid-pattern", List("Error parsing the XML: Premature end of file."), List()),
```
samanthagottlieb, line 91: Just to clarify - where/how would this error be generated? What is it about the example `rule` in this test that would generate this specific error?

samanthagottlieb (approved):
Looks good to me! Have tested locally and it works as expected 👍

samanthagottlieb:
Forgot to mention that CI is still failing - seemingly because `RuleManagerSpec.scala isn't formatted properly`

jonathonherbert:
Thanks @samanthagottlieb! Good spot, I'd forgotten to push my latest changes.

apps/rule-manager/test/db/RuleManagerSpec.scala
jonathonherbert: I think this is happening because there's no `pattern` defined in the rule – so as well as giving an error that the field is required, the validation also complains that there's nothing to parse.

PR #409, opened 2023-08-14
Discard draft changes

samanthagottlieb:
## What does this change?

This PR adds a 'Discard unpublished changes' button when a Live rule has unpublished changes. It replaces the 'Has unpublished changes' warning text with a button that reverts the changes.

Currently, when a user makes changes to a Live rule, a warning appears saying the rule 'Has unpublished changes'. These changes are saved to the draft rule. However, it's not possible for the user to see which changes have been made since the latest live rule, nor is it possible to revert these changes. Say a cat walks over your keyboard as you have a rule open - this button will help you undo the work of their paws!

When a user clicks to 'Discard unpublished changes', the draft rule will be updated with properties from the latest live rule, reverting it to a previous version of the rule.

The only property of the draft rule that will be different from the previous state of the draft rule is the `updatedAt` time. We can't retrieve the `updatedAt` time of the draft rule's previous state at the time of publication - this field will have been overwritten by making changes to the draft rule and this property won't necessary match the `createdAt` time of the latest live rule. So instead the `updatedAt` property of the reverted rule will be set to match the `updatedAt` time of the latest live rule. A mock `updatedAt` time is used in the test.

Trello card is [here](https://trello.com/c/fzHRxgtw/1526-add-a-discard-draft-changes-butto).

## How to test

The test can be run within `sbt` with `testOnly *RuleManagerSpec -- -z "revertDraftRule"`

To test this as a user, do the following:
- publish a draft rule
- make changes to the live rule - the 'Discard changes' button should appear in the 'RULE STATUS' bar - next to the 'Has unpublished changes' warning
- click the 'Discard changes button' - the rule should revert to the most recently published state of the live rule

### Before changes to live rule
discard changes - before changes

### After changes to live rule
discard changes - after changes


jonathonherbert (approved):
Tested locally, looks great! Lovely to see the test, and the revisionId logic makes lots of sense.

Minor UX point: this is a destructive operation and it's not possible to see what's about to happen to the rule you're working on. Is it worth launching a 'confirm' modal that shows a diff, a bit like the one we see for publication? (Obvs we wouldn't need the 'reason'!)

Screenshot 2023-08-15 at 18 32 45

Just something to follow up if you're interested – we should not block merging this PR.

samanthagottlieb:
> Tested locally, looks great! Lovely to see the test, and the revisionId logic makes lots of sense.
> 
> Minor UX point: this is a destructive operation and it's not possible to see what's about to happen to the rule you're working on. Is it worth launching a 'confirm' modal that shows a diff, a bit like the one we see for publication? (Obvs we wouldn't need the 'reason'!)
> 
> Screenshot 2023-08-15 at 18 32 45
> Just something to follow up if you're interested – we should not block merging this PR.

Yes very good point! I had planned to do this as a follow up PR, thanks for the reminder. Will merge this and create a new card for the diff modal.

PR #411, opened 2023-08-17
Discard rule changes modal

samanthagottlieb:
## What does this change?

This adds a Revert modal that allows the user to review their unpublished changes before discarding them.

Previously, when a live rule has unpublished changes, a `Discard unpublished changes` button would appear in the 'Rule Status' bar - button added [here](https://github.com/guardian/typerighter/pull/409). Clicking this would immediately revert the draft rule to its previous state, before the changes were made. However, the user might not be sure which unpublished changes they are discarding. The Revert modal will show a before and after diff to the user on clicking the `Discard unpublished changes` button, allowing them to review before confirming discard.

## How to test

To test this as a user, do the following:

- publish a draft rule
- make changes to the live rule - the 'Discard unpublished changes' button should appear in the 'RULE STATUS' bar
- click the 'Discard unpublished changes button' - the Revert modal should appear, showing a diff with before and after changes were made
- clicking 'Cancel' should return you back to editing the rule, with the unpublished changes still there
- clicking 'Discard changes' should revert to the most recently published state of the live rule

This can also be tested by running the client app tests with `npm run test`.

### DIFF
--- a/apps/rule-manager/client/src/ts/components/RuleForm.tsx
+++ b/apps/rule-manager/client/src/ts/components/RuleForm.tsx
@@ -22,6 +22,7 @@ import { LineBreak } from './LineBreak';
 import { CategorySelector } from './CategorySelector';
 import { TagsSelector } from './TagsSelector';
 import { RuleFormSection } from './RuleFormSection';
+import { RevertModal } from './modals/Revert';
 
 export type PartiallyUpdateRuleData = (
 	partialReplacement: Partial<DraftRule>,
@@ -79,11 +80,13 @@ export const RuleForm = ({
 		unarchiveRule,
 		unpublishRule,
 		ruleStatus,
+		isDiscarding,
 		discardRuleChanges,
 	} = useRule(ruleId);
 	const [ruleFormData, setRuleFormData] = useState(rule?.draft ?? baseForm);
 	const debouncedFormData = useDebouncedValue(ruleFormData, formDebounceMs);
 	const [isReasonModalVisible, setIsReasonModalVisible] = useState(false);
+	const [isRevertModalVisible, setIsRevertModalVisible] = useState(false);
 
 	const partiallyUpdateRuleData: PartiallyUpdateRuleData = (
 		partialReplacement,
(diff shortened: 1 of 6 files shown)


apps/rule-manager/client/src/ts/components/RuleForm.tsx
```diff
@@ -204,12 +207,24 @@ export const RuleForm = ({
...
 		if (!ruleId || ruleStatus !== 'live') {
 			return;
 		}
 
+		setIsRevertModalVisible(true);
+	};
+
+	const discardRuleChangesHandler = async () => {
+		const isDraftOrLive = ruleStatus === 'draft' || ruleStatus === 'live';
+		if (!ruleId || !isDraftOrLive) {
```
jonathonherbert, line 220: Non-blocking: is this really a test for 'archived' status? If so, could we check for that explicitly?

jonathonherbert (approved):
Tested locally, works great. Feels really good to see what's going to happen.

One general non-blocking comment – I was a bit confused to see that the left hand side column represented what would happen if I clicked revert, and the right hand side was what currently existed. It's the other way around in the 'publish' modal, I suspect b/c for left to right readers we associated time moving from left to right (left: what currently is, right: what will happen.)

Might be overthinking it, wonder if lhs as 'before revert' and rhs as 'after revert' would be clearer ↔️ 🤔

apps/rule-manager/client/src/ts/components/RuleForm.tsx
samanthagottlieb: Good point, I suppose it actually needs to check if the status is 'live', as you can only revert a live rule. I'll change this, thanks for spotting.

samanthagottlieb:
> Tested locally, works great. Feels really good to see what's going to happen.
> 
> One general non-blocking comment – I was a bit confused to see that the left hand side column represented what would happen if I clicked revert, and the right hand side was what currently existed. It's the other way around in the 'publish' modal, I suspect b/c for left to right readers we associated time moving from left to right (left: what currently is, right: what will happen.)
> 
> Might be overthinking it, wonder if lhs as 'before revert' and rhs as 'after revert' would be clearer ↔️ 🤔

Yes I had wondered the same, and was on the fence about which way round was better. But that's a good point about moving from left to right - I've changed them around and made the modal text a bit more descriptive about what's going on.

It now looks like this:
Screenshot 2023-08-22 at 14 59 26

PR #412, opened 2023-08-24
Add permissions to appropriate action endpoints

samanthagottlieb:
## What does this change?

This adds a permissions check to several actions that were initially too permissive. A check was added to the `publish`, `refresh` and `refreshDictionaryRules` endpoints as these actions make changes to rule data in the database.

Trello card is [here](https://trello.com/c/WKI2rD8M/1497-add-permissions-to-endpoints-that-dont-currently-have-them).

@rhystmills and I looked at refactoring to combine the authentication checking (currently done using `APIAuthAction`) with authorisation/permissions checking, to create an action `APIAuthActionWithPermission`. We attempted to extend the trait `AbstractApiAuthAction with PlainErrorResponses` and considered overriding the `invokeBlock` method to include permission checking. However it proved trickier than expected to customise to our needs, so we decided to prioritise OKR work and come back to this refactor at a later point.

## How to test

Run Typerighter locally and check you can still publish rules and refresh rules (clicking the red button to destructively pull in rules from the spreadsheet). As you'll already need Composer credentials to run the Typerighter app locally, adding this permissions check should not affect your ability to publish / refresh rules.

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -27,33 +27,41 @@ class RulesController(
     extends PandaAuthController(controllerComponents, config)
     with PermissionsHandler
     with FormHelpers {
-  def refresh = APIAuthAction {
-    val maybeWrittenRules = for {
-      dbRules <- sheetsRuleResource
-        .getRules()
-        .left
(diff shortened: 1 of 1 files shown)


rhystmills (approved):
Looks good. Tested and the methods are working as expected locally 👍

PR #413, opened 2023-08-30
Allow Composer to filter out dictionary matches

rhystmills:
## What does this change?

This PR allows Composer to request a list of matcher categories that it wants included in the checker response. This is the mechanism by which we can, for example, ask for all categories except dictionary rules, in order to exclude them in Composer (if the user toggles them off).

It does this by: 
- Making sure that our `DictionaryRule`s have the category 'Collins Dictionary' by setting the `CATEGORY` in `MorfologikCollinsSpellerRule`.
-  Changing `getApplicableCategories` to return all categories when it is given an empty Set. Until now, an empty list of categories passed to the `checkStream` endpoint gave an empty `Some(Set.empty)` rather than `None` - which meant `getApplicableCategories` would say there were no applicable categories. Basically, if we pass an empty set of categories from this branch, it will consider all categories valid. 

It also stops dictionary rules with multiple suggestions coming through as orange 'Review' matches when they should be be red 'Amend' matches.

## How to test

Test with https://github.com/guardian/flexible-content/pull/4391/files, which will require you to use https://github.com/guardian/prosemirror-typerighter/pull/257 locally.

However, this PR should be good to merge independently of the other two, as long as the last one merged is the flexible-content PR.

- Requesting all categories except dictionary rules returns the expected result
- Request an empty list of categories returns results from all matchers

### DIFF
--- a/apps/checker/app/services/MatcherPool.scala
+++ b/apps/checker/app/services/MatcherPool.scala
@@ -374,9 +374,8 @@ class MatcherPool(
   /** Get the categories to apply to this check. If no categories are supplied, default to all
     * available categories.
     */
-  private def getApplicableCategories(query: Check) = (query.categoryIds, query.excludeCategoryIds) match {
-    case (None, None)      => getCurrentCategories.map(_.id)
-    case (Some(ids), None) => ids
-    case (None, Some(ids)) => getCurrentCategories.map(_.id).diff(ids)
+  private def getApplicableCategories(query: Check) = query.categoryIds match {
+    case None      => getCurrentCategories.map(_.id)
+    case Some(ids) => ids
   }
 }
--- a/apps/checker/app/services/MatcherPool.scala
+++ b/apps/checker/app/services/MatcherPool.scala
@@ -374,8 +374,9 @@ class MatcherPool(
   /** Get the categories to apply to this check. If no categories are supplied, default to all
     * available categories.
     */
-  private def getApplicableCategories(query: Check) = query.categoryIds match {
-    case None      => getCurrentCategories.map(_.id)
-    case Some(ids) => ids
+  private def getApplicableCategories(query: Check) = (query.categoryIds, query.excludeCategoryIds) match {
+    case (None, None)      => getCurrentCategories.map(_.id)
+    case (Some(ids), None) => ids
+    case (None, Some(ids)) => getCurrentCategories.map(_.id).diff(ids)
   }
 }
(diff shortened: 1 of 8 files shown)


apps/checker/app/services/MatcherPool.scala
```diff
@@ -375,7 +375,7 @@ class MatcherPool(
     * available categories.
```
samanthagottlieb, line 375: This comment needs to be changed to reflect how this now works.

apps/checker/app/services/MatcherPool.scala
```diff
@@ -375,7 +375,7 @@ class MatcherPool(
...
     */
   private def getApplicableCategories(query: Check) = query.categoryIds match {
-    case None      => getCurrentCategories.map(_.id)
-    case Some(ids) => ids
+    case Some(ids) if ids.size > 0 => ids
+    case _                         => getCurrentCategories.map(_.id)
```
samanthagottlieb, line 379: I wonder if leaving `categoryIds` as `undefined` if there aren't any suggestions (rather than assigning it to an [empty array](https://github.com/guardian/prosemirror-typerighter/blob/7fe8acffce01092c6f19ed7568a5fca7bc1d3c19/src/ts/components/Match.tsx#L50) in `prosemirror-typerighter`) would make the logic here more straightforward - we wouldn't have to check if the `ids.size` was bigger than 0. However it does feel more explicit to set `suggestions` to an empty array if there aren't any, rather than leaving it as `undefined`.

rhystmills: We could definitely do that in `prosemirror-typerighter` - but I still think the expected behaviour in Typerighter should be to return all categories if the user requests `[]` - no-one will ever want to get no matches back. 

To me the existing pattern match felt misleading:
`Some(ids)` versus `None` felt like it should mean - they give me category ids and I use them, or they give me no ids, and I instead use all categories

Instead it was -
- They give me Some(ids), I use the ids given
- They give me Some(Set.empty), I use no categories (as they asked for a Set of no ids)
- They give me None - I include all categories

apps/checker/app/services/MatcherPool.scala
rhystmills: I think this is how it works as of this PR - if I ask for an empty list of categories, it defaults to all.

Previously, if I asked for an empty list of categories, it would give me no categories.

rhystmills: I think the person who wrote that code may have assumed that an empty list of categories would be transformed to `None` by the play framework

apps/checker/app/services/MatcherPool.scala
samanthagottlieb: Agreed! Ignore me, getting categories and suggestions confused - I blame the heat.

simonbyford: Also agreed!

simonbyford (approved):
I've got this working locally and everything looks good to me!

PR #416, opened 2023-08-31
Show error message when composer credentials are missing

simonbyford:
## What does this change?

Show an error message to the setup script when composer credentials are missing so the failure is less cryptic. 

## How to test

Run the setup script (`./script/setup`) with or without composer credentials and note the relevant output.


rhystmills (approved):
Tested locally with and without credentials and works as expected 👍

PR #417, opened 2023-08-31
Update instructions for running Typerighter within Composer locally

samanthagottlieb:
## What does this change?

This updates the outdated instructions on how to run Typerighter from within Composer locally.

## How to test

- Get Typerighter and Composer running locally
- Create an article in Composer and add some text to it
- Follow the instructions in the doc to integrate the two and run the Typerighter checker

You should be able to integrate and successfully run Typerighter on a Composer article.


rhystmills (approved):
Much better 👍

PR #418, opened 2023-08-31
Only display the total rules per category in the Checker rules page, not all the rules

rhystmills:
With the 250,000+ dictionary rules, serialising all the rules to HTML uses all the available memory when we load the `/rules` page in the Checker service. The Rule Manager shows us all rules already, and we don't want to solve this problem in both apps.

Instead, only show the total rules per category in the Checker. This should give us confidence that the correct number of rules are published for each category (e.g. subcategories of style guide rules, language tool rules, dictionary rule).

image

In the future it may be a good idea to include some kind of artefact id in the Checker, in order to verify that a new artefact from the Rule Manager has been properly ingested by the Checker.

## How to test

1. Run the Checker service according to the instructions in the readme.
2. Visit [the rules page](https://checker.typerighter.local.dev-gutools.co.uk/rules).
3. If there is a table with total rules per category there, including dictionary rules (probably reported as 'Typos'), this branch is working as expected.
	1. If not, run the Rule Manager service, make sure dictionary rules have been created (by hitting the `api/refreshDictionary` endpoint). Then check the `/rules` page in the Checker again.

### DIFF
--- a/apps/checker/app/views/rules.scala.html
+++ b/apps/checker/app/views/rules.scala.html
@@ -27,56 +27,13 @@ rulesIngested: Option[Int] = None, errors: List[String] = List.empty)(implicit r
                     </tr>
                 </thead>
                 <tbody>
-                @for(matcher <- matchers) {
+                @for(matcher <- matchers.sortBy(matcher => 0 - matcher.getRules().size)) {
                     <tr>
                         <td>
                         @matcher.getType()
                         </td>
                         <td>@matcher.getCategories().map(_.id).mkString(", ")</td>
-                        <td>@matcher.getRules().size</td>
-                    </tr>
-                }
-                </tbody>
-            </table>
-        </div>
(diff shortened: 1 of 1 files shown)


apps/checker/app/views/rules.scala.html
```diff
@@ -27,56 +27,13 @@ <h5>Matchers (@matchers.size)</h5>
                     </tr>
                 </thead>
                 <tbody>
-                @for(matcher <- matchers) {
+                @for(matcher <- matchers.sortBy(matcher => 0 - matcher.getRules().size)) {
```
jonathonherbert, line 30:
```suggestion
                @for(matcher <- matchers.sortBy(0 - _.getRules().size)) {
```

(arguably less readable, please feel free to ignore!)

jonathonherbert (approved):
Tested locally, LGTM!

PR #419, opened 2023-08-31
Add note about permissions

simonbyford:
## What does this change?

Add note about permissions to the docs.


Georges-GNM (approved):
Good note 🎵

PR #420, opened 2023-08-31
Add full file path for composer backend properties file

samanthagottlieb:
## What does this change?

This edits the composer integration doc, adding a `~` to the file path where one has to change the typerighter url, making it clear where this file lives.


simonbyford (approved):
~

PR #425, opened 2023-09-01
Filter out title case matches from dictionary rule matches

rhystmills:
## What does this change?

This PR filters out Dictionary rule matches that match against Title Case phrases.

Currently, most false positive spellcheck errors are for proper nouns - usually non-english names and places.

This PR filters out such matches before they are sent to Composer (though the DictionaryMatcher LanguageTool instance still generates such matches).

It also adds some tests which try to cover as many edges cases as possible.

### In action in Composer:
image

## How to test

Run locally:
- Composer, with config pointing to local Typerighter Checker
- Typerighter Checker service
- Typerighter Rule Manager service
- Make sure you have dictionary rules published from the rule manager
- Use Typerighter in an article in your local checker.
- Title Case words and phrases should not be spellchecked

### DIFF
--- a/apps/checker/test/scala/matchers/DictionaryMatcher.scala
+++ b/apps/checker/test/scala/matchers/DictionaryMatcher.scala
@@ -1,0 +1,62 @@
+package matchers
+
+import com.gu.typerighter.model.{Category, DictionaryRule}
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class DictionaryMatcherTest extends AsyncFlatSpec with Matchers {
+  val dictionaryRules = List(
+    DictionaryRule("1", "one", Category("fakeId", "fakeCategory")),
+    DictionaryRule("2", "two", Category("fakeId", "fakeCategory")),
+    DictionaryRule("3", "three", Category("fakeId", "fakeCategory"))
+  )
+  val matcher = new DictionaryMatcher(dictionaryRules)
+
+  "isTitleCase" should "match a single title case word" in {
+    val regexDoesMatch = matcher.isTitleCase("Word")
+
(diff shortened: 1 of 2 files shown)


apps/checker/test/scala/matchers/DictionaryMatcher.scala
```diff
@@ -0,0 +1,62 @@
...
+    regexDoesMatch shouldBe true
+  }
+
+  "isTitleCase" should "match title case words containing accents" in {
+    val regexDoesMatch = matcher.isTitleCase("Él Düderino")
+
+    regexDoesMatch shouldBe true
+  }
+
+  "isTitleCase" should "not match mixed case word sequences" in {
```
jonathonherbert, line 57: Ooh, why not?

jonathonherbert (commented):
Ran tests locally, works as expected.

One question – we'd expect corrections at the start of a sentence to be startcased, e.g. `... end of sentence. Strat of next sentence` would have correction `Start`. Would this PR account for sentence starts?

There's a bit of prior art for detecting and managing sentence starts in RegexMatcher if that's relevant.

apps/checker/test/scala/matchers/DictionaryMatcher.scala
rhystmills: It's a bit arbitrary - I just wanted it to clearly only match Title Case phrases, and e.g. a lower case word among the capitalised first-letter words would be an exception to that.

In practice, I've only seen single word dictionary matches so far so this is unlikely to manifest itself in our matches.

jonathonherbert:
@rhystmills, do we still need this after #443?

rhystmills:
> @rhystmills, do we still need this after #443?

We do not - now closed.

PR #426, opened 2023-09-06
Use EuiDataGrid to display rules, adding sorting and server-side pagination

jonathonherbert:
## What does this change?

Uses EuiDataGrid to display rules, adding sorting and server-side pagination. EuiDataGrid offers a few nice features, including inspecting individual cells, and a few customisation options. You can see a few of them in the gif below.

The main advantage of moving to server-side pagination is that we'll be able to cope with the larger rulesets provided by our dictionary, whilst still having all our rules in the same place for searching, batch editing, etc.

euidatagrid

In moving search to the server side, its behaviour has changed slightly – it now searches across description, pattern, source, and replacement for partial matches, whereas EUI's default behaviour would search across the entire dataset. Searching with our full ruleset of ~300k should feel very snappy – I get response times of ~50ms locally.

## Dev notes

- Currently the page size is 200 – enough to let us navigate, but not so many that EUI's scroll bar is unreasonably small. Happy to change!
- I've moved from ScalikeJDBC's DSL to SQL interpolation for this query – I originally wrote it in the DSL, but it added complexity and required several type casts. Hopefully what we've got here is more readable whilst enabling us to reuse ordering and search clauses.
- I originally planned to split this PR into two, but ultimately I think there are some advantages to shipping as a single unit: as inspecting the results of searching and filtering is easier with a UI, and because the search API changes, we'd have to work backwards a bit to ship the first, server-side part of the work without disrupting the current interface. I'm happy to revisit this decision if splitting would be better.

## How to test

I've added tests for sorting and filtering on the server side.

Running locally or in CODE, have a play:
- Searching partial queries
- Sorting by 1 or more columns
- Playing with column order, column visibility, and display options
- Single selection and rule editing
- Multi selection

Does it all behave as expected?


apps/rule-manager/app/model/PaginatedResponse.scala
```diff
@@ -0,0 +1,19 @@
...
+    data: List[Data],
+    pageSize: Int,
+    page: Int,
+    pages: Int,
+    total: Int
+)
+
+object PaginatedResponse {
+  @nowarn
+  implicit def format[Data](implicit _underlying: Format[Data]): Format[PaginatedResponse[Data]] =
```
andrew-nowak, line 17: grid does this with the slightly terser 
```suggestion
  implicit def format[Data: Format]: Format[PaginatedResponse[Data]] =
```
( I hadn't internalised exactly how this syntax works before - TIL this is a [_context bound_](https://docs.scala-lang.org/scala3/book/ca-context-bounds.html) which Scala compiler will convert into essentially the same as your version - but without the "unused implicit" warning :) )

jonathonherbert: So much nicer – thank you!

rhystmills:
Really fast, looks great and feels good to use. Searching is miles better than what we have now, and great to be able to sort and hide columns without investing a bunch of engineering time.

I got an out-of-memory error while unpublishing some fields, which would be worth looking into. Most useful log I saw was this, which is not especially helpful:

```
{"@timestamp":"2023-09-11T10:40:06.875+01:00","@version":"1","message":"SQL execution failed (Reason: Ran out of memory retrieving query results.):\n\n   select r.rule_type, r.pattern, r.replacement, r.category, r.description, r.notes, r.external_id, r.force_red_rule, r.advisory_rule, r.created_at, r.created_by, r.updated_at, r.updated_by, r.revision_id, r.rule_order, r.reason, r.is_active, COALESCE(ARRAY_AGG(rt.tag_id) FILTER (WHERE rt.tag_id IS NOT NULL), '{}') AS tags from rules_live r left join rule_tag_live rt on r.external_id = rt.rule_external_id and r.revision_id = rt.rule_revision_id where r.is_active = true group by r.rule_type, r.pattern, r.replacement, r.category, r.description, r.notes, r.external_id, r.force_red_rule, r.advisory_rule, r.created_at, r.created_by, r.updated_at, r.updated_by, r.revision_id, r.rule_order, r.reason, r.is_active order by r.rule_order\n","logger_name":"scalikejdbc.StatementExecutor$$anon$1","thread_name":"application-akka.actor.default-dispatcher-10","level":"ERROR","level_value":40000,"application.home":"/Users/rhys_mills/code/typerighter/apps/rule-manager"}
```
**More subjective, and less important:**
I find EUI's multi-field sorting mildly confusing - I wonder if it would be simpler to just overwrite an existing sort in the frontend with the last sort chosen. Did you have any specific use cases for multi-sort in mind?

I also wonder if we should exclude empty entries when we sort - at the moment sorting Description `z-a` shows many rows with no description before we see the z's. 

Open to thoughts on all this though. Amazing work and looking forward to seeing it released.

jonathonherbert:
@rhystmills thanks for this! We've taken a look out-of-bound and we think the OoM issues are current behaviour with large rulesets – will be worth optimising in another PR.

re: sort, yes, it takes a bit of getting used to. I can't see an easy way to turn it off in EUI at present, but perhaps we could follow up in another PR (or submit something to EUI if we really didn't like it.)

rhystmills (dismissed):

rhystmills (approved):

PR #428, opened 2023-09-11
Add breadcrumbs

jonathonherbert:
## What does this change?

Adds breadcrumbs to the rule manager.

|Before|After|
|-|-|
|Screenshot 2023-09-11 at 16 41 34|Screenshot 2023-09-11 at 16 41 09|

## Why?

This change anticipates adding additional routes to the rule manager as part of our rule testing work – we'll want a route for `rules/:id/test`.

It also anticipates us moving selection state into the route. At the moment it's not possible to deep-link an individual rule or a rule selection, but it'd be good if we could move that state into the URL to allow users to share links to rules.

## Dev notes

I've changed the way our routing is structured to enable the use of `useMatches`, which requires a static route declaration and is [recommended in the docs](https://reactrouter.com/en/main/routers/create-browser-router) for later versions of react-router.

I've also lifted our page layout out of our routes to prevent repetition, and shuffled the page components into `./pages` to better reflect the component structure.

## How to test

Running locally, or in CODE, navigate around the app. Do the breadcrumbs behave as expected? Is the layout working as it should?

### DIFF
--- a/apps/rule-manager/client/src/ts/components/PageNotFound.tsx
+++ b/apps/rule-manager/client/src/ts/components/PageNotFound.tsx
@@ -54,17 +54,7 @@ const TypeyArrow = styled.img`
 
 export const PageNotFound = () => {
 	return (
-		<>
-			<EuiFlexItem
-				grow={false}
-				css={css`
-					padding-bottom: 20px;
-				`}
-			>
-				<EuiTitle>
-					<h1>404 - Page not found</h1>
-				</EuiTitle>
-			</EuiFlexItem>
+		<div>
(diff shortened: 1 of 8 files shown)


rhystmills:
image

It feels a bit odd that both the breadcrumbs here link to the same page (the one you're on) - should this just be `Home`?

rhystmills (approved):
Looks good! As discussed out of band - the big scary buttons have disappeared, but otherwise looking good.

jonathonherbert:
Thanks for spotting! 74c2837f2352d6e007b01fa502d2a9d5ff06af45

PR #429, opened 2023-09-12
Allow checker to write the collins.dict file

rhystmills:
## What does this change?

In CODE and PROD we are currently unable to write the`collins.dict` file due to a permissions error. This PR makes sure `typerighter-checker` can write to it.


jonathonherbert (approved):
Tested in CODE ✅

PR #430, opened 2023-09-12
Use context for tags

jonathonherbert:
## What does this change?

Moves tags to `Context`, reducing the amount of prop drilling necessary to supply tags to the rules table, tag table, and single/multiple edit forms. Necessary to move to a parameterless Route for rule editing.

## How to test

Running locally or in CODE, examine the four components above. Do tags appear correctly in the rules table, single/batch edit forms, and tag table?


samanthagottlieb:
> I checked out `main` and the same is happening, so at least we know it's not as a result of this PR, but thought I'd point out anyway - I'll create a ticket to fix.

Small follow up to this - I couldn't recreate the issue when the Source of the tag is 'Style guide and names' or 'Names' - only when the Source is 'Check this' (I haven't tested all Sources).

Card is [here](https://trello.com/c/NiThfFsM/1711-fix-tag-reordering-bug-in-rule-manager-diff-modals).

samanthagottlieb (approved):
Tested locally and the rules table, tag table, and single/multiple edit forms work as expected 👍

jonathonherbert:
Thanks for spotting the issue, @samanthagottlieb!

PR #432, opened 2023-09-13
A few UI tweaks/fixes for the rules table

jonathonherbert:
## What does this change?

A few tweaks to the rules table:

- Fixes a crash when users select a rule that's not on page 1 😅 
- Adds a summary to indicate how many rules are available for the current search, displayed, and selected (if more than one rule is selected)
- Sorts the rules by ruleType ascending, putting regex rules first, to make the initial rules page look a bit more sensible (we will have ruleType in a column in another PR, which will make this clearer to users) – at the moment we surface LTRuleCore rules first, which results in a very blank first set of rules for PROD datasets (LTRuleCore rules do not have pattern, description etc.)

|No selection, no search|With selection|With search|
|-|-|-|
|Screenshot 2023-09-13 at 16 01 38|Screenshot 2023-09-13 at 15 32 53|Screenshot 2023-09-13 at 15 33 36|

## How to test

Have a play in local or CODE. Useful? Comments?

### DIFF
--- a/apps/rule-manager/app/db/DbRuleDraft.scala
+++ b/apps/rule-manager/app/db/DbRuleDraft.scala
@@ -245,7 +245,7 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
         }
       }
 
-      sqls"ORDER BY ${sqls.join(orderStmts, sqls",")}"
+      sqls"ORDER BY ${sqls.join(orderStmts, sqls",")}, ${rd.ruleType} ASC"
     } else sqls.empty
 
     val searchClause = maybeWord
--- a/apps/rule-manager/client/src/ts/components/hooks/useRules.ts
+++ b/apps/rule-manager/client/src/ts/components/hooks/useRules.ts
@@ -12,7 +12,6 @@ export type PaginatedResponse<Data> = {
 
 export type PaginatedRuleData = {
 	data: DraftRule[];
-	loadedRules: Set<number>;
 	pageSize: number;
 	total: number;
 };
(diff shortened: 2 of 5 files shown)


rhystmills (commented):
image

The first page is still showing the blank rows we thought were LanguageTool Core rules first - the other changes are great and working as expected 👍

jonathonherbert:
Thanks for spotting, mistakenly reverted in [4881555](https://github.com/guardian/typerighter/pull/432/commits/4881555f3703d665b78fbd0ebb53b3ab41df2489), force-pushed to correct.

simonbyford (commented):
https://github.com/guardian/typerighter/pull/432/commits/9a4df92cd8c9aed114d3e84a98c014e6d09a1839 contains some changes to `rule-audit-client` - was this deliberate?

jonathonherbert:
@simonbyford yuck, thanks for spotting! Just removed those.

simonbyford (approved):
Nice, this is looking good! 🚀 

One small regression I noticed is that selected rows after page 1 are no longer highlighted correctly.

Screenshot 2023-09-25 at 12 07 16

One quick fix would be add `pagination.pageIndex * pagination.pageSize` to this line...

https://github.com/guardian/typerighter/pull/432/files#diff-91ee043de564685aef9af51b599d85f2b3848673333d0c430026629c5a334c40R340

...as this would compensate for pagination.

But I'll leave this with you if you think it's worth fixing now. Possibly highlighting requires a wider rethink?

jonathonherbert:
Great spot, added in a4784b63389e4cd5bff6f7da5554fef8dd01beee – multi-selection is a bit unloved ATM but I think the core state handling will remain the same even as we start to use routes for state when editing individual rules 👍

simonbyford:
> Great spot, added in [a4784b6](https://github.com/guardian/typerighter/commit/a4784b63389e4cd5bff6f7da5554fef8dd01beee) – multi-selection is a bit unloved ATM but I think the core state handling will remain the same even as we start to use routes for state when editing individual rules 👍

Noice, that did the job!

This is unrelated but I've just noticed something else a bit odd. Try alternating between page 1 and page 5 and note which row gets loaded at the top each time. When I try this locally, the top row changes each time. Could this point to something being wrong with the sorting? Super weirdly, this doesn't happen when alternating between page 1 and page 2. 

Happy to pair if any of that doesn't make sense!

PR #433, opened 2023-09-14
Allow list of words not to publish

rhystmills:
## What does this change?

This PR adds support for a list of words to no publish as official dictionary words.

## Why?
Our first audience for the Collins dictionary will be writing in British English, as this is the primary language variant for the majority of our staff. We will want to avoid publishing (for example) American regional variants of words, so that our Dictionary matcher will not recognise them as valid words, and suggest changing to the British English variant.

(In the future we will want to add support for American and Australian regional spellings, but that's not in scope for our pre-MVP work). 

Being a record of the English language, the Collins dictionary also contains a lot of offensive words that we wouldn't want to suggest as autocorrect options, so we may want to disallow them too ([LanguageTool should prevent a few of these](https://github.com/languagetool-org/languagetool/blob/48e8fcb11a68d079a43ba05c403a332feb2c4e69/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/AbstractEnglishSpellerRule.java#L129), but it's by no means an exhaustive list). 

## How?

We will include a list of "words to not publish" in S3, to be used alongside our existing config. I've also modified the setup script (and other local setup-related files) to pull the same file from the dev S3, if it exists, and use it locally.

The JSON is expected to be an array of objects containing a tag name, and a list of associated words. A very short file might look like this:

```json
[
   {
      "tag":"American region spelling",
      "words":[
         "color"
      ]
   },
   {
      "tag":"Offensive word",
      "words":[
         "cretin"
      ]
   }
]
```

We decided to keep the list it in S3 rather than having it here in version control, to avoid the prospect of our repo housing a long list of incredibly offensive words. The tags will help us keep track of excluded words, and perhaps allow a filtering mechanism for consumers in the future (e.g. opt into Americanisms).

## How to test

1. Run Typerighter Rule Manager locally according to the instructions in the readme.
2. Run the setup script, or otherwise add the above to `~/.gu/typerighter/words-to-not-publish.json`
3. Enable the only feature switch in the top right dropdown of the manager
4. Click the scary red button that reloads dictionary rules
5. Search for dictionary words that should not be live according to your config. Are they in draft? Do they have the expected tags?


apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -416,33 +416,55 @@ object RuleManager extends Loggable {
...
-        ruleType = "dictionary",
-        pattern = Some(wordAndIndex._1),
-        category = Some("Collins Dictionary"),
-        ignore = false,
-        user = "Collins Dictionary",
-        ruleOrder = initialRuleOrder + wordAndIndex._2,
-        externalId = None
+    val dictionaryRules = words.zipWithIndex
+      .map(word => {
+        val tagId = wordsToNotPublish.find(taggedWord => taggedWord.word == word) match {
```
jonathonherbert, line 441: Does this definitely do what's expected? The lhs is `Wordtag` and the rhs is `(String, Int)`

apps/rule-manager/app/service/RuleManager.scala
```diff
@@ -416,33 +416,55 @@ object RuleManager extends Loggable {
...
+          ruleType = "dictionary",
+          pattern = Some(wordIndexAndTag._1),
+          category = Some("Collins Dictionary"),
+          ignore = false,
+          user = "Collins Dictionary",
+          ruleOrder = initialRuleOrder + wordIndexAndTag._2,
+          externalId = None,
+          tags = wordIndexAndTag._3
+        )
       )
```
jonathonherbert, line 459: A few things –
 - Do we need two maps here?
 - I'd be tempted to destructure the tuple to make it clear which bit is which.
 - Underscore notation makes things more concise, but you may prefer a more explicit syntax
 - We could use `flatMap` and `flatten` rather than `.get` to avoid unhappy paths throwing errors (although perhaps it's better to fail fast and loudly in this case!)

e.g.

```scala
    val dictionaryRules = words.zipWithIndex
      .map { case (word, index) =>
        val tagId = wordsToNotPublish.find(_.word == word) match {
          case Some(wordTag) =>
            List(availableTags.find(_.name == wordTag.tag).flatMap(_.id)).flatten
          case _ => Nil
        }
        DbRuleDraft.withUser(
          id = None,
          ruleType = "dictionary",
          pattern = Some(word),
          category = Some("Collins Dictionary"),
          ignore = false,
          user = "Collins Dictionary",
          ruleOrder = initialRuleOrder + index,
          externalId = None,
          tags = tagId
        )
      }
```

jonathonherbert (commented):
Nice! A few code comments, just firing it up locally to test.

apps/rule-manager/app/service/RuleManager.scala
rhystmills: Discussed out of band - it wasn't working.

apps/rule-manager/app/service/RuleManager.scala
rhystmills: All good suggestions, will include!

script/setup
```diff
@@ -10,7 +10,18 @@ printf "✅ Pre-commit hooks set up\n"
 printf "⚙️  Downloading dictionary for local use\n"
 {
     aws s3 cp --profile composer --region eu-west-1 s3://typerighter-app-dev/dictionary/typerighter-dictionary.xml ~/.gu/typerighter/typerighter-dictionary.xml &&
+    aws s3 cp --profile composer --region eu-west-1 s3://typerighter-app-dev/dictionary/words-to-not-publish.json ~/.gu/typerighter/words-to-not-publish.json &&
```
jonathonherbert, line 13: IIUC at the moment we download everything twice! 🪞🪞 

```suggestion
```

script/setup
```diff
@@ -10,7 +10,18 @@ printf "✅ Pre-commit hooks set up\n"
...
     printf "✅ Dictionary downloaded\n"
 } || {
     printf "❌ Dictionary download failed. Are your Composer credentials up to date?\n"
 }
+
+printf "⚙️  Downloading 'words to not publish' for local use\n"
+{
+    aws s3 cp --profile composer --region eu-west-1 s3://typerighter-app-dev/dictionary/typerighter-dictionary.xml ~/.gu/typerighter/typerighter-dictionary.xml &&
```
jonathonherbert, line 21: As above.

```suggestion
```

jonathonherbert (approved):
Tested locally, can see draft rules and tags! 🎉 

(It took me a while to test, as ingesting the entire dictionary was ooming the service – might take a look at improving this)

PR #434, opened 2023-09-15
Add rule check UI

jonathonherbert:
NB: depends on #445 – please review that first!

## What does this change?

A first pass at a rule checking UI.

The intent is to provide a rule creator with quick feedback by checking lots of content with their rule. By showing these matches, it should be possible to surface false positives without resorting to other tooling, or publishing the rule and checking content in our CMS.

test-rule

A few notes:
- I've added it on a separate page, accessible via the 'wrench' icon in the rules table, per #445. My feeling is that this page should probably just be the edit mode, and we should remove the sidebar for rule editing, but I haven't done that in this PR as it didn't feel necessary to get this work in. WDYT?
- Aware that the matches could do with more behaviour – correct/review/amend semantics, see suggestions, etc., as well as links to the relevant article, etc. More to come.

## How to test

Have a play with a few rules, changing the CAPI query and the rule pattern. Does the testing behave as expected?

You'll need to run both checker and manager services for this to work – `./script/start` should do the job.

### DIFF
--- a/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
+++ b/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
@@ -27,7 +27,7 @@ export const Rule = () => {
 				{rule?.ruleType !== 'dictionary' ? (
 					<TestRule pattern={testPattern} />
 				) : (
-					<h2>You cannot yet test dictionary rules.</h2>
+					<h2>Testing is not currently supported for dictionary rules.</h2>
 				)}
 			</EuiFlexItem>
 		</EuiFlexGroup>
--- a/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
+++ b/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
@@ -11,6 +11,7 @@ export const Rule = () => {
 	const { id } = useParams() as { id: string };
 	const ruleId = id === newRuleId ? undefined : parseInt(id);
 	const ruleHooks = useRule(ruleId);
+	const rule = ruleHooks.rule?.draft;
 	const testPattern = ruleHooks.rule?.draft.pattern;
 	const navigate = useNavigate();
 	return (
@@ -23,7 +24,11 @@ export const Rule = () => {
 				/>
 			</EuiFlexItem>
 			<EuiFlexItem grow={2}>
-				<TestRule pattern={testPattern} />
+				{rule?.ruleType !== 'dictionary' ? (
+					<TestRule pattern={testPattern} />
+				) : (
+					<h2>You cannot yet test dictionary rules.</h2>
+				)}
 			</EuiFlexItem>
 		</EuiFlexGroup>
 	);
--- a/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
+++ b/apps/rule-manager/client/src/ts/components/pages/Rule.tsx
@@ -9,9 +9,9 @@ export const newRuleId = 'new-rule';
 
 export const Rule = () => {
 	const { id } = useParams() as { id: string };
-  const ruleId = id === newRuleId ? undefined : parseInt(id);
-  const ruleHooks = useRule(ruleId);
-  const testPattern = ruleHooks.rule?.draft.pattern;
+	const ruleId = id === newRuleId ? undefined : parseInt(id);
+	const ruleHooks = useRule(ruleId);
+	const testPattern = ruleHooks.rule?.draft.pattern;
 	const navigate = useNavigate();
 	return (
 		<EuiFlexGroup style={{ height: '100%' }}>
@@ -19,12 +19,12 @@ export const Rule = () => {
 				<RuleForm
 					ruleId={ruleId}
 					onClose={() => navigate('/')}
-          {...ruleHooks}
+					{...ruleHooks}
 				/>
 			</EuiFlexItem>
 			<EuiFlexItem grow={2}>
-         <TestRule pattern={testPattern} />
-      </EuiFlexItem>
+				<TestRule pattern={testPattern} />
+			</EuiFlexItem>
 		</EuiFlexGroup>
 	);
 };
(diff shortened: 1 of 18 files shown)


samanthagottlieb:
This is exciting! A few initial comments/questions.

I've just tried testing this locally and am getting 'No matches found for this search.'. Do I need to be running CAPI locally to test?

Having watched the video, I'm a bit unclear on what the two search behaviours are doing. There's a search happening when you type into the 'Pattern' box, which I'm guessing is searching for that pattern in all articles in CAPI. But you can also narrow down the articles to only ones that contain the word(s) you're trying to match on, is that right? Can you only search for words in the CAPI query box (rather than regexes)?

Also, what is the goal of the testing - is it for your pattern to cause the word(s) that you've searched for on the right to become bold (i.e. confirming that your regex is successfully picking up the pattern you intend it to)? I think if the flow of the page was the other way round, i.e. finding articles with your target words, then testing your regex against the articles, this would feel more intuitive to me. Maybe another good reason to have a separate rule editing/testing page!

apps/rule-manager/client/src/ts/components/RuleContent.tsx
```diff
@@ -146,6 +150,18 @@ export const RuleContent = ({
...
 					label={<Label text="Pattern" required={true} />}
+					labelAppend={
+						<Link
+							to={
+								isTestingRule
+									? `/rule/${ruleData?.draft.id}`
+									: `/rule/${ruleData?.draft.id}/test`
+							}
+						>
+							<EuiIcon type="bug" />
```
samanthagottlieb, line 161: Would the bug icon look better on the right side, to be consistent with the 'Preview' and 'Discard unpublished changes' icons above? Up to you though!

jonathonherbert: Good suggestion – resolving as the UI has changed since this was asked, the relevant bit no longer exists!

jonathonherbert:
This PR has been reworked to put the testing on a separate page. 

Thanks for the feedback, @samanthagottlieb! 

_> I've just tried testing this locally and am getting 'No matches found for this search.'. Do I need to be running CAPI locally to test?_

You'll need to be running both the checker and the manager service at once, as we use the checker service to test the rules. `./script/start` should do the job.

_> Having watched the video, I'm a bit unclear on what the two search behaviours are doing. There's a search happening when you type into the 'Pattern' box, which I'm guessing is searching for that pattern in all articles in CAPI. But you can also narrow down the articles to only ones that contain the word(s) you're trying to match on, is that right? Can you only search for words in the CAPI query box (rather than regexes)?_

I've reworked the UI in the hope that this a bit clearer! You're absolutely right – the intent is to check lots of content with the rule you are working on. By default, we just get what's new in CAPI. You can narrow down the search with a CAPI query.

At the moment, the CAPI query is limited to a plaintext search (no regexes.) This is good for spotting individual instances of a term.

_> Also, what is the goal of the testing - is it for your pattern to cause the word(s) that you've searched for on the right to become bold (i.e. confirming that your regex is successfully picking up the pattern you intend it to)? I think if the flow of the page was the other way round, i.e. finding articles with your target words, then testing your regex against the articles, this would feel more intuitive to me. Maybe another good reason to have a separate rule editing/testing page!_

This is interesting! Might be best talked through out of band? I'll mention on chat.

apps/rule-manager/client/src/ts/components/pages/Rule.tsx
```diff
@@ -2,21 +2,34 @@ import React from 'react';
...
 					onClose={() => navigate('/')}
+					{...ruleHooks}
 				/>
 			</EuiFlexItem>
-			<EuiFlexItem grow={2}></EuiFlexItem>
+			<EuiFlexItem grow={2}>
+				{rule?.ruleType !== 'dictionary' ? (
+					<TestRule pattern={testPattern} />
+				) : (
+					<h2>You cannot yet test dictionary rules.</h2>
```
rhystmills, line 30:
```suggestion
					<h2>Testing is not currently supported for dictionary rules.</h2>
```

rhystmills (commented):
Amazing work! Is this accidentally the nicest way to query CAPI out there?

Overall I think it's a fantastic implementation of the feature, and a great bit of engineering.

I have nothing to add about the code itself, which looks great, but I do have a few thoughts about UI and messaging:

**On the inline menu:**

I think the inline edit menu remains a convenient way to edit rules, and I'd be keen to keep it. I recall Ana had a concept of a single view that would 'pan' between the table view and the testing view, which would be nice to get to. Expect this would be a decent chunk of UI work so wouldn't suggest adding it in this PR, but perhaps one for the future - and I think your solution in this PR is the right one for now.

But, thinking out loud - I'm imagining something like the following, where you pan between the outlined views by clicking some sort of testing button in the inline edit menu:

view 1
view 2

**On messaging:**
We're using some fairly technical language at the moment and might be making some assumptions about what the user would understand. I'd propose expanding the tooltip to say something like the following:

>By default, this tool will search the first 50 pages of Guardian content from CAPI, our content API, containing [X] pieces of content. By adding a CAPI search query, you can narrow down the search to content more likely to include your match. For example, to search the following rule, for 'However' without a comma:
>(?<=\. )However(?!\,)
>…you may want to search for your match only in articles containing 'However', rather than just using 50 pages of the most recent CAPI content.

I'd also change the search placeholder text to say "a search phrase for CAPI" instead of "a CAPI query" to be a bit simpler to understand.

**Some visual tweaks**:
- The tooltip icon looks a bit tiny on my screen
- I reckon the beaker icon from EUI is a good one for testing
Screenshot 2023-10-20 at 10 12 00

jonathonherbert:
_> I'm imagining something like the following, where you pan between the outlined views by clicking some sort of testing button in the inline edit menu:_

This crossed my mine also 😀 totally doable, perhaps in a separate PR?

rhystmills (approved):
Looks good to me 👍

PR #435, opened 2023-09-18
Remove audit app

jonathonherbert:
## What does this change?

Removes the audit app, introduced as a little experiment way back in  https://github.com/guardian/typerighter/pull/44.

We've got a story for testing rules pending in the rule manager client (e.g. #389), so this code is no longer needed, and removing it will speed up CI and reduce our dependency surface area.

## How to test

- The app should build and run as before
- There should be no 'audit' link in the header of the checker UI:

Screenshot 2023-09-18 at 10 47 37


simonbyford:
You absolutely love to see it

Screenshot 2023-09-22 at 15 16 40

Tested on CODE (as my local environment is broken at the moment) and everything looks great, nice one!

simonbyford (approved):

PR #436, opened 2023-09-19
Add new property `groupKey` to `RuleMatch`

simonbyford:
## What does this change?

Adds a new property `groupKey` to a rule match which indicates how the rule should be grouped when it produces matches. 

This is needed because currently all Collins Dictionary rules are confusingly grouped together by the Typewriter client as they all share the same rule ID (`MORFOLOGIK_RULE_COLLINS`). This motivated the creation of a new field to control grouping. 

- For non-dictionary rules, it's just the rule ID.
- For dictionary rules, it's a combination of the rule ID and the matched text.

Publishing this change in isolation won't have any effect, but once this PR is merged the the Typerighter client can then be updated to use `groupKey` to group rules in the sidebar, instead of `ruleId`.

## How to test

Run typerighter and composer locally, ensuring composer is pointing at the local typerighter instance as per [the docs](https://github.com/guardian/typerighter/blob/main/docs/03-composer-integration.md). Ensure the Collins Dictionary feature switch is enabled. Create an article, open the browser network tab, make a spelling mistake and click "Check document". When you inspect the response from the HTTP request, the JSON should contain the new property `groupKey`. Something like:

```
{
    "matches": [
            {
                "rule": {
                    "id": "MORFOLOGIK_RULE_COLLINS",
                    "category": {
                        "id": "Collins Dictionary",
                        "name": "Collins Dictionary"
                    },
                    ...
                },
                ...
                "groupKey": "MORFOLOGIK_RULE_COLLINS-sherbert"
            }
        ]
...
}
```

### DIFF
--- a/apps/checker/app/matchers/DictionaryMatcher.scala
+++ b/apps/checker/app/matchers/DictionaryMatcher.scala
@@ -36,7 +36,16 @@ class DictionaryMatcher(
   override def check(
       request: MatcherRequest
   )(implicit ec: ExecutionContext) = {
-    matcher.check(request)
+    // groupKey is used to control how rules are grouped in the client when they produces matches.
+    // This is needed for dictionary matches as they all share a common rule ID (MORFOLOGIK_RULE_COLLINS)
+    // groupKeys for dictionary matches have the format `MORFOLOGIK_RULE_COLLINS-{matchedText}`
+    matcher
+      .check(request)
+      .map(ruleMatches =>
+        ruleMatches.map(ruleMatch =>
+          ruleMatch.copy(groupKey = Some(ruleMatch.rule.id + '-' + ruleMatch.matchedText))
+        )
+      )
   }
 
   def getCategories() = instance.getAllActiveRules.asScala.toList.map { rule =>
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/fixtures/RuleMatchFixtures.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/fixtures/RuleMatchFixtures.scala
@@ -16,6 +16,7 @@ object RuleMatchFixtures {
     subsequentText = "",
     matchedText = "placeholder text",
     message = "placeholder message",
-    matchContext = "[placeholder text]"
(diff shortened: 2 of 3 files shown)


jonathonherbert (commented):
Code looks great. I'd recommend something to document why we're doing this for dictionary matches – this could be a comment at the point at which we're mapping to add the group key, and/or a test for the dictionary matcher to ensure that we're getting back what we expect.

Blocked on one test is failing in `RuleTestingSpec.scala` for an obscure reason – AFAICS it looks like we're tipped some vendor code over an optimisation threshold or some other condition, and it's decided to return a `Vector` rather than a `List` for the `result` variable assigned on L161.

This is possible because we've declared the return type of to be `Seq`, which is a supertype of both `List` and `Vector`.

Calling `toList` on the result should fix – happy to pair, and if anyone else knows precisely why this has happened (may be Akka specific)  I'd be interested to hear it!

simonbyford:
> Code looks great. I'd recommend something to document why we're doing this for dictionary matches – this could be a comment at the point at which we're mapping to add the group key, and/or a test for the dictionary matcher to ensure that we're getting back what we expect.
> 
> Blocked on one test is failing in `RuleTestingSpec.scala` for an obscure reason – AFAICS it looks like we're tipped some vendor code over an optimisation threshold or some other condition, and it's decided to return a `Vector` rather than a `List` for the `result` variable assigned on L161.
> 
> This is possible because we've declared the return type of to be `Seq`, which is a supertype of both `List` and `Vector`.
> 
> Calling `toList` on the result should fix – happy to pair, and if anyone else knows precisely why this has happened (may be Akka specific) I'd be interested to hear it!

Thanks for the comment and the green tick @jonathonherbert. The issue with the test seemed to be that `groupKey` was missing from a text fixture which makes sense.

I've added a small comment to document the change and definitely agree a test would be great. Perhaps we could pair on adding one?

jonathonherbert:
Oh nice, I saw List->Vector but missed the property change! Glad it works well. As mentioned out of band, we can definitely ship this now and circle back with a test.

PR #383, opened 2023-07-14
Add HMAC authentication to checkSingle endpoint

jonathonherbert:
_PR 2/3 to to fulfil the workflow described in https://github.com/guardian/typerighter/discussions/344._

## What does this change?

Adds HMAC authentication to the checker service so we can authenticate HMAC headers for the `checkSingle` endpoint.

## Implementation notes

To do this, we needed to use `pan-domain-auth-play_2-8` in both the manager and checker. Previously, the checker was using `pan-domain-auth-validation`, as it only needs to verify cookies, and not create them. `panda-hmac-play_2-8` depends on `pan-domain-auth-play_2-8`, and we need hmac in our checker, so this PR expands the capabilities of the checker service slightly, allowing it to mint pan-domain-auth cookies.

This has a few benefits:
- testing the checker service no longer requires us getting a cookie from another `.local` service
- we can DRY up the code for managing authentication between services, which means that despite adding another authentication method in this PR, the diff removes lines.

I've leaned into that with a refactor that moves our various controller configurations into a single class, `PandaAuthController`, which configures both sorts of authentication – both services require the HMAC library and config, as both reading and writing HMAC headers requires access to panda secrets, so there's no unnecessary configuration.

## How to test

- All the current endpoints should authenticate as expected. Running locally or in CODE, you should be able to access both the manager and checker service as before – but only when authenticated. Try removing your cookies for each service and hitting an authenticated endpoint (everything that isn't a healthcheck.) You should go through an oAuth flow and sign in as usual.
- The HMAC authentication should work as expected. Testing this is a little like #363, but ... hard mode! Following the steps in that PR, we'll also need to sign our request with HMAC headers.

We can do that by sending a request as per #363, but we'll need to sign the request with HMAC headers. I've added a script to do that in Postman below. The secret is in AWS Secrets Manager in the usual account – speak to me if it's not clear how to get hold of it. You'll need the secret for the `DEV` stage.

Alternatively, in the next PR in the series, the rule manager will use this authentication to talk to the checker service directly, so you could use that PR to verify _this_ PR's behaviour. 🙃 

<details>
<summary>Postman script</summary>

```javascript
const URL = require("url");

const uri = pm.request.url.toString();
const secret = "changeme";

const httpDate = new Date().toUTCString();
const path = new URL.parse(uri).path;
const stringToSign = `${httpDate}\n${path}`;
const stringToSignBytes = CryptoJS.enc.Utf8.parse(stringToSign);
[...]

PR #437, opened 2023-09-19
Reinstate auth  grace period

jonathonherbert:
## What does this change?

Reinstates an auth grace period, introduced in https://github.com/guardian/typerighter/pull/66 and lost I think in the refactor in https://github.com/guardian/typerighter/pull/383.

See the original PR for the rationale.

## How to test

Bit of a long'un – leave a Composer tab open for longer than an hour, and attempt to run a check. Does it still work?

### DIFF
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/controllers/PandaAuthController.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/controllers/PandaAuthController.scala
@@ -24,6 +24,8 @@ abstract class PandaAuthController(
     with HMACAuthActions
     with BaseController
     with Loggable {
+  override val apiGracePeriod: Long = 2 * 60 * 60 * 10000 // two hours worth of milliseconds
+
   override def validateUser(authedUser: AuthenticatedUser): Boolean = {
     log.info(s"Validating user $authedUser")
     PanDomain.guardianValidation(authedUser)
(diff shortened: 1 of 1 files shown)


samanthagottlieb (approved):
Looks good to me - tested by leaving Composer tab open for over an hour and ran a check - it worked 🎉

Out of interest, without this expiry extension, do you know how/where the auth period is defined? I was surprised by how quickly it expired during the demo.

apps/common-lib/src/main/scala/com/gu/typerighter/controllers/PandaAuthController.scala
```diff
@@ -24,6 +24,8 @@ abstract class PandaAuthController(
     with HMACAuthActions
     with BaseController
     with Loggable {
+  override val apiGracePeriod: Long = 2 * 60 * 60 * 10000 // two hours worth of milliseconds
```
davidfurey, line 27: I think you should be multiplying by one thousand rather than ten thousand here

samanthagottlieb: Yikes well spotted - sorry I missed this!

jonathonherbert: Thanks for spotting!
```suggestion
  override val apiGracePeriod: Long = 2 * 60 * 60 * 1000 // two hours worth of milliseconds
```

PR #438, opened 2023-09-22
Fix edit rule button

simonbyford:
## What does this change?

Previously, the "edit rule" button was broken from page 2 and onwards:

before

This was happening because it attempts to lookup the rule by `id` in a sparse array (one padded with many `null` values). From page 2 and onwards, the array looks something like:

```
[
   null,
   null,
   null,
   null,
   null,
   ...
   {
      id: 1,
      ruleType: "dictionary",
      pattern: "(c)"
      ...
   }
]
```

So it freaks out when `id` doesn't exist as a property. We can use optional chaining to avoid this. 

## How to test

Run the rule manager locally, navigate to page 2, and attempt to edit any rule. It should work.

## Images

after


samanthagottlieb (approved):
Well spotted! 👍

PR #440, opened 2023-09-29
Remove whitespace-only patterns from dictionary rules

jonathonherbert:
## What does this change?

We've a couple of dictionary rules that are imported but not published after import. That's because they just contain whitespace, and as such they fail the form validation checks (`nonEmptyString`).

The whitespace is caused by the formatted XML.

This PR aims to be a bit more specific in which nodes it retrieves words from, and as a fallback removes any empty words from the XML output, so dictionary rules are published immediately on import.

## How to test

The current DEV XML will produce this sort of whitespace, so running locally, try importing the dictionary. Previous to this PR, it would fail. It should now succeed and all dictionary rules should be marked live.

### DIFF
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/rules/Dictionary.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/rules/Dictionary.scala
@@ -5,9 +5,10 @@ import scala.xml.{Elem, Node}
 object Dictionary {
   def lemmaOrInflListToText(node: Node): List[String] = {
     node match {
-      case node if node.label == "lemma"     => List(node.text)
-      case node if node.label == "infl_list" => node.child.toList.map(infl => infl.text)
-      case _                                 => Nil
+      case node if node.label == "lemma" => List(node.text)
+      case node if node.label == "infl_list" =>
+        node.child.filter(_.label == "infl").map(_.text).toList
+      case _ => Nil
     }
   }
 
@@ -25,7 +26,6 @@ object Dictionary {
       lemmaOrInflList <- lemmaOrInfl.toList
       word <- lemmaOrInflListToText(lemmaOrInflList)
     } yield word
-    words.distinct.filterNot(_ == "")
+    words.distinct.filterNot(_.filterNot(_.isWhitespace) == "")
   }
-
 }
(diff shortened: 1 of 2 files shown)


apps/common-lib/src/main/scala/com/gu/typerighter/rules/Dictionary.scala
```diff
@@ -25,7 +26,6 @@ object Dictionary {
       lemmaOrInflList <- lemmaOrInfl.toList
       word <- lemmaOrInflListToText(lemmaOrInflList)
     } yield word
-    words.distinct.filterNot(_ == "")
+    words.distinct.filterNot(_.filterNot(_.isWhitespace) == "")
```
davidfurey, line 29: Would `words.distinct.filterNot(_.forall(_.isWhitespace))` be more obvious?

apps/common-lib/src/test/scala/com/gu/typerighter/rules/DictionaryTest.scala
```diff
@@ -19,6 +19,22 @@ class DictionaryTest extends AnyFlatSpec with Matchers {
     expected should be(actual)
   }
 
+  "dictionaryXmlToWordList" should "not include strings that only contain whitespace" in {
```
davidfurey, line 22: Is it worth adding a test for being more specific in which nodes it retrieves words from?

apps/common-lib/src/main/scala/com/gu/typerighter/rules/Dictionary.scala
jonathonherbert:
```suggestion
    words.distinct.filterNot(_.forall(_.isWhitespace))
```

Much nicer, ta!

apps/common-lib/src/test/scala/com/gu/typerighter/rules/DictionaryTest.scala
jonathonherbert: Good idea, 6219a18bd17972b8cec69022b08bb028661e79c9.

rhystmills (approved):
Looks good to me!

PR #443, opened 2023-10-05
Filter named entities from dictionary matches

simonbyford:
## What does this change?

Right now, proper nouns cause false positives in Typerighter because they're not found in Collins Dictionary. This creates a lot of noise and erodes trust in the tool.

Screenshot 2023-10-05 at 09 09 37

This PR attempts to improve this by identifying named entities from text being checked and excluding them from dictionary matches. We use the named-entity recognition (NER) library [OpenNLP](https://opennlp.apache.org/) to achieve this. 

To start with, we use 3 models: people, organisations and locations. There is a full list of models here:

https://opennlp.sourceforge.net/models-1.5/

In the future, we might wish to use the model developed internally by the data science team as it was trained on Guardian content and will likely do a better job:

https://github.com/guardian/data-science-ner-service

We just went with OpenNLP as the simplest proof of concept.

## How to test

Run typerighter and composer locally, ensuring composer is pointing at the local typerighter instance as per [the docs](https://github.com/guardian/typerighter/blob/main/docs/03-composer-integration.md). Ensure the Collins Dictionary feature switch is enabled. Try adding some proper nouns and see if they are flagged as misspellings or not.

Fewer false positives caused by proper nouns in text.

There is a risk that genuinely misspelled words might be incorrectly identified as named-entities and therefore excluded from spell-checking (i.e. false negatives).

## Images

TODO


apps/checker/app/matchers/DictionaryMatcher.scala
```diff
@@ -33,17 +33,28 @@ class DictionaryMatcher(
...
+    val namedEntities = request.blocks.flatMap((block) => entityHelper.getAllEntitiesFromText(block.text))
+
     matcher
       .check(request)
       .map(ruleMatches =>
-        ruleMatches.map(ruleMatch =>
-          ruleMatch.copy(groupKey = Some(ruleMatch.rule.id + '-' + ruleMatch.matchedText))
+        ruleMatches
+          // Remove matches which correspond to named entities. This should reduce the number of false-positives
+          // caused by pronouns
```
simonbyford, line 51:
```suggestion
          // caused by proper nouns
```

apps/checker/app/services/EntityHelper.scala
```diff
@@ -0,0 +1,63 @@
...
+  val locationFinder = new NameFinderME(locationModel)
+
+  def getAllEntitiesFromText(text: String): List[EntityInText] = {
+    getEntitiesFromText(text, "person") ++
+      getEntitiesFromText(text, "organization") ++
+      getEntitiesFromText(text, "location")
+  }
+
+  def getEntitiesFromText(text: String, model: String): List[EntityInText] = {
+    val finder = model match {
```
jonathonherbert, line 34: Non-blocking: rather than pattern match on a `String` type here, which will never be exhaustive (hence the match that throws), I'd be tempted to put all the finders in a list, and iterate across that – something like

```scala
// In the class constructor
val allFinders = List(personFinder, ...)

// In `getEntitiesFromText`
finders.foldLeft(List.empty[EntityInText])((finder, entities) => {
  val entitySpans = finder.find(tokenSpans.map(_.getCoveredText(text).toString))
  // ... accumulate entities
})
```

feel like that'll remove an opportunity for an error to be thrown and might be fewer lines – WDYT?

simonbyford: Good shout, this didn't feel great before, I've made that change here: https://github.com/guardian/typerighter/pull/443/commits/e0fa022c13e8829eb17b174a3510accd6298a380

It does mean we lose the `model` attribute but I think that's fine - we weren't using it for anything anyway.

apps/checker/app/matchers/DictionaryMatcher.scala
```diff
@@ -33,18 +33,35 @@ class DictionaryMatcher(
...
   override def check(
       request: MatcherRequest
-  )(implicit ec: ExecutionContext) = {
-    // groupKey is used to control how rules are grouped in the client when they produces matches.
-    // This is needed for dictionary matches as they all share a common rule ID (MORFOLOGIK_RULE_COLLINS)
-    // groupKeys for dictionary matches have the format `MORFOLOGIK_RULE_COLLINS-{matchedText}`
+  )(implicit ec: ExecutionContext): Future[List[RuleMatch]] = {
+    val entityHelper = new EntityHelper()
+    val namedEntities =
+      request.blocks.flatMap((block) => entityHelper.getEntitiesFromText(block.text))
```
jonathonherbert, line 50: We'll need to include the offset of the current block when marking entity ranges – entity ranges will begin at the start of their block, but match ranges will be relative to the entire request, so if the block doesn't start at 0, entities and ranges won't match up!

I can see this when checking on CODE – every prosemirror document begins at position `2`, so entities and matches don't line up. (Also should be verifiable with a test for a block that starts at a nonzero position.)

This should be as simple as 

```
TextRange(
  from = from + block.from,
  to = to + block.from
)
```

Passing an `offset` into `getEntitiesFromText` would be one way of solving.

jonathonherbert:
This looks good! With the offset problem corrected, there's definitely a reduction of noise:

|Before|After|
|-|-|
|Screenshot 2023-10-05 at 12 30 32|Screenshot 2023-10-05 at 12 30 59|

Having said that, there are two problems, one of which might block this approach:
- OpenNLP is very slow! It takes ~7-8s to check the above three paragraphs locally. That's on a snappy M1!
- The model is not super accurate. You can see it misses 'Bojangles', 'Medgar Evans' and 'Bojangles' above.

Interestingly, the vanilla spaCy model gets pretty much everything right, and is much quicker – this comes back in ~100-120ms:

```json
[
  { "label": "PERSON", "text": "Lisa Simone", "start": 0, "end": 11 },
  { "label": "PERSON", "text": "Nina Simone", "start": 113, "end": 124 },
  { "label": "ORG", "text": "Broadway", "start": 464, "end": 472 },
  { "label": "PERSON", "text": "Nina", "start": 892, "end": 896 },
  { "label": "PERSON", "text": "Lisa Simone", "start": 935, "end": 946 },
  { "label": "PERSON", "text": "Nina Simone’s", "start": 1108, "end": 1121 },
  { "label": "PERSON", "text": "Medgar Evers", "start": 1327, "end": 1339 },
  {
    "label": "PERSON",
    "text": "Lorraine Hansberry",
    "start": 1457,
    "end": 1475
  },
  { "label": "PERSON", "text": "Weldon Irvine", "start": 1492, "end": 1505 },
  { "label": "PERSON", "text": "Bojangles", "start": 1744, "end": 1753 }
]
```

Wonder whether it's worth spiking a call out to ner.gutools service to see how it performs? Should be fairly quick to try.

apps/checker/app/matchers/DictionaryMatcher.scala
jonathonherbert, line 50: Worth noting that this code will run the entity finder first, and then the check.

Might be nice to do that work in parallel. We can wrap blocking code in a `Future` to run it in another thread:
```suggestion
      Future { request.blocks.flatMap((block) => entityHelper.getEntitiesFromText(block.text)) }
```

We can then run both futures and perhaps use a for-comprehension to wait for them both before continuing:

```scala
for {
  result1 <- future1
  result2 <- future2
} yield {
  // do something w/ result1, result2
}
```

If we migrate to making HTTP calls to an external service, we'll likely get that for free, so perhaps we don't need to optimise yet.

simonbyford: Good catch, thank you. I've attempted a fix here: 

https://github.com/guardian/typerighter/pull/443/commits/765e886db51e4048ab869d38077765f859bd999d

Will test on CODE shortly to verify

apps/checker/app/services/EntityHelper.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  val organizationInputFile: InputStream =
+    getClass.getClassLoader.getResourceAsStream("resources/opennlp/en-ner-organization.bin")
+  val organizationModel = new TokenNameFinderModel(organizationInputFile)
+  val organizationFinder = new NameFinderME(organizationModel)
+
+  val locationInputFile: InputStream =
+    getClass.getClassLoader.getResourceAsStream("resources/opennlp/en-ner-location.bin")
+  val locationModel = new TokenNameFinderModel(locationInputFile)
+  val locationFinder = new NameFinderME(locationModel)
```
simonbyford, line 25: Could the slowness be attributable to this block being run on every check, instead of just once?

apps/checker/app/matchers/DictionaryMatcher.scala
```diff
@@ -33,18 +33,40 @@ class DictionaryMatcher(
...
+  }
+
   override def check(
       request: MatcherRequest
-  )(implicit ec: ExecutionContext) = {
-    // groupKey is used to control how rules are grouped in the client when they produces matches.
-    // This is needed for dictionary matches as they all share a common rule ID (MORFOLOGIK_RULE_COLLINS)
-    // groupKeys for dictionary matches have the format `MORFOLOGIK_RULE_COLLINS-{matchedText}`
+  )(implicit ec: ExecutionContext): Future[List[RuleMatch]] = {
+    val entityHelper = new EntityHelper()
```
simonbyford, line 48: [See below comment](https://github.com/guardian/typerighter/pull/443/files#r1350109866)

apps/checker/app/services/EntityHelper.scala
jonathonherbert: Ooh – yes. How does it perform when we instantiate the helper once per application?

(I meant to check this, and didn't get to it. Thanks for spotting.)

apps/checker/app/matchers/DictionaryMatcher.scala
simonbyford: Great suggestion, I've made the change: https://github.com/guardian/typerighter/pull/443/commits/19244d3ff8de44f45c22413886a7609026241445

apps/checker/app/services/EntityHelper.scala
simonbyford: I've made this change: 

https://github.com/guardian/typerighter/pull/443/commits/560ab776c4bcf089b926308da07adffe2d394f7b

Seems promising based off logging some timestamps, construction does indeed take much longer than finding entities! But I haven't yet tested on CODE

jonathonherbert (approved):
Tested locally and CODE, works as expected. 
Performance is significantly improved – we can keep an eye on the response times in PROD. The model is still not good, but does filter out some useful matches!

Logging the entities might be useful, so we can see just how much noise this technique is reducing in PROD – I've added an example commit on another branch @ 30d9bb97e7a160b2e5a12806946163cd2c579cca. WDYT?

simonbyford:
> Tested locally and CODE, works as expected. Performance is significantly improved – we can keep an eye on the response times in PROD. The model is still not good, but does filter out some useful matches!

Thanks!

> Logging the entities might be useful, so we can see just how much noise this technique is reducing in PROD – I've added an example commit on another branch @ [30d9bb9](https://github.com/guardian/typerighter/commit/30d9bb97e7a160b2e5a12806946163cd2c579cca). WDYT?

I like it very much, have cherry-picked onto this branch

PR #445, opened 2023-10-09
New route for editing rules

jonathonherbert:
## What does this change?

Adds client-side routing for rule editing, with page that lets you edit the rule. It's a bit blank yet, but there's space for testing to be added by #434. 

Screenshot 2023-10-09 at 11 12 01

There's no way to access it from the UI, yet, but you can get to it via the URL.

## How to test

Try navigating to the URL in the screenshot, filling in the relevant rule ID as necessary. The client should look like the screenshot above.


rhystmills (approved):
Worked as intended for me 👍

rhystmills (approved):
Worked as intended for me 👍

PR #448, opened 2023-10-12
Correct asset paths

jonathonherbert:
## What does this change?

#445 introduced a routing change that broke the site assets. This PR fixes.

## How to test

Run locally, and in CODE. You should see:

- [x] the site content load as normal (JS, CSS and asset files like fonts are correct)
- [ ] the favicon appear as normal

Tested:

- [x] locally
- [x] on CODE

NB: I had to check in Firefox for the favicon to appear correctly – there appears to be some weird favicon caching in Chrome! Checking the path in the markup manually should verify if it's not appearing straight away in Chrome and you're unsure.


samanthagottlieb:
The favicon is appearing like this for me locally:
Screenshot 2023-10-12 at 15 21 44

However it looks fine on CODE.

PR #452, opened 2023-10-17
Make artefact creation and read more efficient

rhystmills:
## What does this change?

The Checker and Rule Manager services frequently crash due to `OutOfMemory` errors, especially locally. Watching the service run while measuring memory usage, we have seen that memory spikes correlate with the point at which the artefact is serialised to JSON by the Rule Manager and read by the Checker.

At this point, the services potentially have quite a lot in memory - the `List[CheckerRule]` - containing >250,000 entries, the `JsValue` representation of that list, and the string representation of that JSON (~50mb in size).

This PR writes the artefact as newline-delineated JSON, iterating through the rules and writing bytes to the artefact for each individual rule, so that we don't need to have as many objects in memory at any one time, and giving the JVM garbage collector more opportunities to reclaim memory.

This should resolve the `OutOfMemory` errors on artefact write. Time from publishing in the Rule Manager to receiving the new match in Composer seems a lot faster - from around 14 seconds in testing, but I've been unable to benchmark against existing times because the Checker crashes too often on older branches for me to measure on those branches.

The PR should be backwards compatible, checking first for the new (newline delineated JSON) artefact and falling back to the old (standard JSON) artefact if it's not found.

(We should be able to remove this in the future and move only to the new style).

## How to test

1. It's a good idea to deploy to CODE and check the backwards compatibility. Make sure the new artefact (`typerighter-rules-seq.json`) is not present in the Typerighter CODE S3 bucket for a fair test of the backwards compatibility (you may need to manually delete it from S3 via the AWS console). Does the deploy complete?
2. Once successfully deployed, create a new regex rule and publish it. 
3. Does it show up in Composer CODE against matching text in a timely manner? (It should).
4. Does the checker service crash with an `OutOfMemory` error? (It shouldn't).

1. No OutOfMemory errors on artefact publication
2. Faster time from publication in the Rule Manager to matches appearing in Composer

### DIFF
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/lib/JsonHelpers.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/lib/JsonHelpers.scala
@@ -14,6 +14,9 @@ object JsonHelpers {
   def toJsonSeq[T](serializable: T)(implicit tjs: Writes[T]) =
     Json.toJson(serializable).toString() + recordSeparatorChar
 
+  def toNewlineDeliniatedJson[T](serializable: T)(implicit tjs: Writes[T]) =
+    Json.toJson(serializable).toString() + "\n"
+
   val JsonSeqFraming: Flow[ByteString, ByteString, NotUsed] =
     Framing.delimiter(ByteString(recordSeparatorChar), Int.MaxValue, true)
 }
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/lib/JsonHelpers.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/lib/JsonHelpers.scala
@@ -14,7 +14,7 @@ object JsonHelpers {
   def toJsonSeq[T](serializable: T)(implicit tjs: Writes[T]) =
     Json.toJson(serializable).toString() + recordSeparatorChar
 
-  def toNewlineDeliniatedJson[T](serializable: T)(implicit tjs: Writes[T]) =
+  def toNewlineDelineatedJson[T](serializable: T)(implicit tjs: Writes[T]) =
     Json.toJson(serializable).toString() + "\n"
 
   val JsonSeqFraming: Flow[ByteString, ByteString, NotUsed] =
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
@@ -5,11 +5,11 @@ import com.amazonaws.services.s3.model.{ObjectMetadata, PutObjectRequest}
 import com.gu.typerighter.lib.JsonHelpers
 import com.gu.typerighter.model.{CheckerRule, CheckerRuleResource}
 import play.api.Logging
-import play.api.libs.json.{Json}
+import play.api.libs.json.Json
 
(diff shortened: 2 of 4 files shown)


apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
```diff
@@ -2,47 +2,93 @@ package com.gu.typerighter.rules
...
+      try {
+        val rules = s3.getObject(bucketName, RULES_KEY)
+        val lastModified = rules.getObjectMetadata.getLastModified
+        val rulesStream = rules.getObjectContent()
+        val rulesArray = ArrayBuffer[CheckerRule]()
+        val reader = new BufferedReader(new InputStreamReader(rulesStream))
+        reader.lines.forEach(line => {
+          rulesArray += Json.parse(line).as[CheckerRule]
+        })
+        val rulesList = rulesArray.toList
```
davidfurey, line 48:
```suggestion
        val rulesList = Source.fromInputStream(rulesStream).getLines().map(line => {
          Json.parse(line).as[CheckerRule]
        }).toList
```

apps/common-lib/src/main/scala/com/gu/typerighter/lib/JsonHelpers.scala
```diff
@@ -14,6 +14,9 @@ object JsonHelpers {
   def toJsonSeq[T](serializable: T)(implicit tjs: Writes[T]) =
     Json.toJson(serializable).toString() + recordSeparatorChar
 
+  def toNewlineDeliniatedJson[T](serializable: T)(implicit tjs: Writes[T]) =
```
jonathonherbert, line 17: Nit: typos in a spellchecker mean double irony points 

```suggestion
  def toNewlineDelineatedJson[T](serializable: T)(implicit tjs: Writes[T]) =
```

jonathonherbert (approved):
Tested in CODE w/ #449, rule publish/unpublish works as expected, and I couldn't get the CODE box to fall over 💪

apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
```diff
@@ -2,47 +2,93 @@ package com.gu.typerighter.rules
...
+  private val RULES_KEY = s"$stage/rules/typerighter-rules-seq.json"
+  private val LEGACY_RULES_KEY = s"$stage/rules/typerighter-rules.json"
 
   def putRules(ruleResource: CheckerRuleResource): Either[Exception, Unit] = {
-    val ruleJson = Json.toJson(ruleResource)
-    val bytes = Json.toBytes(ruleJson)
-
+    val ruleJsonBytes = ArrayBuffer[Byte]();
+    ruleResource.rules.foreach(rule => {
+      ruleJsonBytes ++= JsonHelpers.toNewlineDeliniatedJson(rule).getBytes()
```
rhystmills, line 21:
```suggestion
      ruleJsonBytes ++= JsonHelpers.toNewlineDelineatedJson(rule).getBytes()
```

apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
rhystmills: Cheers, now implemented!

PR #453, opened 2023-10-20
Use Guardian NER service for entity recognition

simonbyford:
Co-authored-by: @rhystmills 

## What does this change?

Our [first attempt](https://github.com/guardian/typerighter/pull/443) at NER (named entity recognition) has resulted in a [reduction in noise](https://github.com/guardian/typerighter/pull/443#issuecomment-1748715074) but many proper nouns remain unidentified and are therefore flagged as typos by typerighter.

peckham

This PR swaps our NER implementation from the [OpenNLP](https://opennlp.apache.org/) library to the Guardian's own NER service, which is accessed via a REST API. This results in a significant further reduction in noise. However, there are a few risks associated with this approach, as outlined below.

If we decide to go ahead with this approach, we can strip out the OpenNLP implementation in a subsequent PR. 

## How to test

Either run this branch locally (in conjunction with composer) or deploy it to CODE. Paste in some text and see how many proper nouns are flagged as spelling mistakes, compared with doing the same thing on `main`. There are a few test articles below.

There are 3 main risks:

1) As before, there is a risk with any NER solution that genuinely misspelled words might be incorrectly identified as named-entities and therefore excluded from spell-checking (i.e. false negatives).

2) There is a risk that adding this integration degrades or overwhelms the Guardian NER service because of the increased traffic to the API. This would be bad because it is used by other production systems. We can mitigate this by working closely with the data science team to monitor the service.

3) There is a risk that introducing a network request increases the latency of the checker and this feels obvious to the user.

Fewer proper nouns being flagged as spelling mistakes. No significant hit to the usability of the tool.

### [Ukraine benchmarking article](https://composer.code.dev-gutools.co.uk/content/652cf0ba8f081a136f3a40f5)

|Before|After|
|-|-|
| Screenshot 2023-10-23 at 10 17 14 | Screenshot 2023-10-23 at 10 17 36 |

### [Israel benchmarking article](https://composer.code.dev-gutools.co.uk/content/652cf03d8f081a136f3a40f3)

|Before|After|
|-|-|
| Screenshot 2023-10-23 at 10 16 58 | Screenshot 2023-10-23 at 10 17 06 |

### [Art benchmarking article](https://composer.code.dev-gutools.co.uk/content/65363bf68f08b60f838f1eed)

|Before|After|
|-|-|
| Screenshot 2023-10-23 at 10 53 06 | Screenshot 2023-10-23 at 10 53 28

## Guardian NER Links

Repo: https://github.com/guardian/data-science-ner-service
Logs: https://logs.gutools.co.uk/s/data-science/app/discover 
PROD Service: https://ner.gutools.co.uk/v1/
CODE Service: https://ner.code.dev-gutools.co.uk/v1/


apps/checker/app/matchers/DictionaryMatcher.scala
```diff
@@ -59,40 +59,45 @@ class DictionaryMatcher(
       entities: List[EntityInText]
   ): Boolean = {
     entities.exists(entity =>
-      entity.range.from == ruleMatch.fromPos && entity.range.to == ruleMatch.toPos
+      entity.range.from <= ruleMatch.fromPos && entity.range.to >= ruleMatch.toPos
     )
   }
 
+  def matchFallsWithinNamedEntity(
```
jonathonherbert, line 66: Is there a reason we wouldn't also consider the position here when lining up found entities with text, as with `matchFallsWithinNamedEntityRange`? It removes the (presumably unlikely, but idk!) possibility of a word being used as a noun in one position, but another part of speech (e.g. verb) in another.

apps/checker/app/matchers/DictionaryMatcher.scala
```diff
@@ -59,40 +59,45 @@ class DictionaryMatcher(
       entities: List[EntityInText]
   ): Boolean = {
     entities.exists(entity =>
-      entity.range.from == ruleMatch.fromPos && entity.range.to == ruleMatch.toPos
+      entity.range.from <= ruleMatch.fromPos && entity.range.to >= ruleMatch.toPos
```
jonathonherbert, line 62: Is this method still needed? (Perhaps, given comment below 😀  – but I think we should only need one?)

apps/checker/test/scala/services/EntityHelperTest.scala
```diff
@@ -1,55 +0,0 @@
-package services
```
jonathonherbert, line 1: IIUC, if we keep the code in `EntityHelper` above, perhaps best to keep the test! Or, we could remove them both?

jonathonherbert (commented):
Unable to test locally at the moment – I can circle back on Thurs unless somebody else gets to this before then 😀 

A few non-blocking comments on the code.

I should add that both the content of this PR description and the results of your testing are _delightful._ 👏 👏

apps/checker/app/services/EntityHelper.scala
```diff
@@ -26,6 +32,45 @@ class EntityHelper() {
...
+
+  object NERResult {
+    implicit val reads: Reads[NERResult] = Json.reads[NERResult]
+  }
+
+  def getEntityResultFromNERService(text: String): Future[Either[Error, List[String]]] = {
+    val url = config.nerApiUrl
+    val key = config.nerApiKey
+    val model = "en_core_web_trf"
+    val entityTypes = List("PERSON", "NORP", "FAC", "LOC", "GPE", "PRODUCT", "EVENT", "WORK_OF_ART")
```
jonathonherbert, line 45: Perhaps we could include "ORG", too? Corporation/org names are often novel.

```suggestion
    val entityTypes = List("PERSON", "NORP", "FAC", "LOC", "GPE", "PRODUCT", "EVENT", "WORK_OF_ART", "ORG")
```

e.g. 

Screenshot 2023-10-26 at 10 16 54

but 

```json
{
    "result": [
        {
            "text": "...",
            "ents": [
                {
                    "label": "ORG",
                    "text": "gal-dem",
                    "start": 0,
                    "end": 7
                }
            ]
        }
    ]
}
```

jonathonherbert (approved):
LGTM, one comment about "ORG" entities might be worth a look, some minor comments but approving to unblock.

rhystmills:
New commits looks good to me 👍

apps/checker/app/services/EntityHelper.scala
```diff
@@ -34,8 +38,11 @@ class EntityHelper(wsClient: WSClient, config: CheckerConfig) {
       )
       .post(body)
       .map { response =>
+        val after = System.currentTimeMillis
         response.status match {
           case 200 => {
+            logger.info(s"Request to ${config.nerApiUrl} succeeded in ${after - before}ms")
```
jonathonherbert, line 44: Might be worth adding markers to this log line so we can measure latency as a separate field in ELK – [here's an example.](https://github.com/guardian/typerighter/blob/2181de7552016cfa3259dbccc58dc69d8f1a269c/apps/checker/app/services/MatcherPool.scala#L51)

apps/checker/app/matchers/DictionaryMatcher.scala
jonathonherbert: Double checking that you've seen this – our approach has changed here, might make perfect sense but I want to make sure I understand why!

simonbyford: Ahhh, I knew there was a reason we used ranges but had forgotten!

We changed it because we thought it might be weird if the same word gets marked as incorrect in one place and not another. It had the further benefit of simplifying the model (passing around `List[String]` instead of `List[EntityInText]`).

However, on reflection I think it makes more sense to use ranges, so I've reverted it here: https://github.com/guardian/typerighter/pull/453/commits/389754863022486ec35f087b4e236f369fba47bc

apps/checker/app/matchers/DictionaryMatcher.scala
simonbyford: We're using it again now!

apps/checker/app/services/EntityHelper.scala
simonbyford: Good catch, I've added it

apps/checker/app/matchers/DictionaryMatcher.scala
jonathonherbert: Gotcha, yeah that's an interesting case – appreciate we're sorta at the mercy of our model here 😅

apps/checker/app/services/EntityHelper.scala
simonbyford: Ah good shout. The `Timer` class which seems to do exactly what we need here. See: https://github.com/guardian/typerighter/pull/453/commits/08dec4c5034e92ede157a9581fbeafd984292620?w=1 (best viewed with whitespace hidden)

jonathonherbert (approved):

PR #454, opened 2023-10-24
Bump CDK dependencies

rhystmills:
## What does this change?

This PR bumps some CDK dependencies in order to resolve a vulnerability in `@guardian/cdk` (derived from `@babel/traverse`). None of the breaking changes in that library from `v50` or `v51` should affect Typerighter.

It also bumps the node version to`16.14.0` - as the previous version was incompatible with those dependencies. 

I tried bumping higher to `18` and `20` and strangely found that one of our Scala tests failed much more regularly (locally and in CI) when I did (`it should "reject work that exceeds its buffer size"...` in `MatcherPoolTest`). Might require some digging when we want to bump Node further.

## How to test

- Run the manager and checker locally according to the instructions in the readme to check that they are compatible with the new (very similar) Node version. Do they build and run as expected? 
- Deploy this branch to CODE. If it works, the build and deploy have worked as intended and we can be happy with the CDK changes.

1. Package builds and deploys as expected.
2. The vulnerability is no longer reported by Snyk after merge.


cdk/package.json
```diff
@@ -12,14 +12,14 @@
...
-    "jest": "^27.1.0",
-    "ts-jest": "^27.0.5",
+    "@guardian/cdk": "51.0.0",
+    "@types/jest": "^29.5.6",
+    "@types/node": "20.8.7",
+    "aws-cdk": "2.100.0",
+    "aws-cdk-lib": "2.100.0",
+    "constructs": "^10.3.0",
+    "jest": "^29.5.0",
+    "ts-jest": "^29.1.0",
```
samanthagottlieb, line 22: Is there a reason you chose to use the caret for `jest` related dependencies, and pin the others? Could be a good idea to pin them all, as per [recommendations](https://github.com/guardian/recommendations/blob/main/dependencies.md#javascript).

samanthagottlieb (approved):
Have tested running the rule manager and checker locally and both work as expected. Just a suggestion to pin dependencies, otherwise looks good!

cdk/package.json
rhystmills: Good point! I've made them all exact versions now.

PR #463, opened 2024-03-20
Bump postgresql lib

davidfurey:
Resolves snyk vulnerability


davidfurey:
Deployed to CODE, rule manager can still view and edit rules

jonathonherbert (approved):

PR #464, opened 2024-03-20
Upgrade to Play 2.9, fix Scala Steward

rtyley:
This upgrade to Play 2.9 was prompted by https://github.com/guardian/scala-steward-public-repos/pull/67 which requires that all projects that want to have Scala Steward updates (like this one) need to be able to _build_ under Java 21 (even if the projects are still _running_ in production on Java 11).

Building under Java 21 [requires](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-21-compatibility-notes) a recent version of sbt, and that leads to a version bump for `scala-xml` that gives version-compatibility errors:

```
[error] 	* org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over {1.2.0, 1.1.1}
[error] 	    +- org.scala-lang:scala-compiler:2.12.18              (depends on 2.1.0)
[error] 	    +- com.typesafe.sbt:sbt-native-packager:1.5.2 (scalaVersion=2.12, sbtVersion=1.0) (depends on 1.1.1)
[error] 	    +- com.typesafe.play:twirl-api_2.12:1.5.1             (depends on 1.2.0)
```

...the nicest way to fix these errors is to upgrade to Play 2.9 or above, so here we are!

https://github.com/guardian/pan-domain-authentication didn't get Play 2.9 artifacts until relatively recently, so upgrading the Panda version was also necessary.

See also:
* https://github.com/guardian/maintaining-scala-projects/issues/1

_(unfortunately the Scala Steward run failing affects everyone, see https://github.com/guardian/scala-steward-public-repos/issues/60 - so I've removed `typerighter` from Scala Steward until this PR is merged)_

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,6 +28,7 @@ val languageToolVersion = "6.0"
 val awsSdkVersion = "1.12.576"
 val capiModelsVersion = "17.5.1"
 val capiClientVersion = "19.2.1"
+val pandaVersion = "3.0.1"
 val circeVersion = "0.14.1"
 val scalikejdbcVersion = scalikejdbc.ScalikejdbcBuildInfo.version
 val scalikejdbcPlayVersion = "2.8.0-scalikejdbc-3.5"
@@ -35,15 +36,6 @@ val appsFolder = "apps"
 
 def javaVersionNumber = {
   IO.read(new File(".java-version"))
-}
-
-def removeStartingOne(javaVersionString: String): String = {
-  val startsWithOne = "^1\\.".r
-  startsWithOne.replaceAllIn(javaVersionString, "")
-}
-
-def parseJavaVersionNumber(javaVersionString: String): String = {
-  removeStartingOne(javaVersionString).split('.').head
 }
 
 val jackson = {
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -126,15 +118,6 @@ def playProject(label: String, projectName: String, domainPrefix: String, devHtt
...
       commonSettings,
-      initialize := {
-        val _ = initialize.value
-        val expectedJavaVersion = parseJavaVersionNumber(javaVersionNumber)
-        val actualJavaVersion = removeStartingOne(sys.props("java.specification.version"))
-        assert(
-          expectedJavaVersion.equals(actualJavaVersion),
-          s"Java $expectedJavaVersion is required for this project. You are using Java $actualJavaVersion."
-        )
-      },
```
rtyley, line 137: We now require that projects be able to _build_ with Java 21 if they're using Scala Steward, even if they aren't _using_ Java 21, so this strict rule here of requiring exactly Java 11 for the build is not helpful. We _could_ have changed it to a greater-than-or-equals condition, but that's already accomplished by the `-release:11` config already present:

https://github.com/guardian/typerighter/blob/b12225dc7eafe1b2b3ed6e4c1103ed9fccc74c49/build.sbt#L11

davidfurey (approved):
LGTM. Any reason not to go straight to Play 3.0?

rtyley:
> Any reason not to go straight to Play 3.0?

I'm just minimising what I need to check before fixing Scala Steward!

PR #465, opened 2024-03-20
Add Prout to know when PRs are deployed or overdue

rtyley:
Adds [Prout](https://www.theguardian.com/info/developer-blog/2015/feb/03/prout-is-your-pull-request-out), to let us know when our changes don't ship as expected, or when they are ready to check in Production!

Some other recents PRs enabling Prout on repos:

* https://github.com/guardian/permissions/pull/174
* https://github.com/guardian/dotcom-rendering/pull/9692

PR #466, opened 2024-03-20
Only use Prout on endpoints without heavy caching

rtyley:
Following on from #465, of these two checkpoint endpoints:

* https://manager.typerighter.gutools.co.uk/healthcheck - goes direct to ELB, no caching
* https://checker.typerighter.gutools.co.uk/healthcheck - is behind Cloudfront, and seems to have 1-hour cache time

...because of the long cache time on Checker, I think it's going to be too confusing to point Prout at it?

image
https://prout-bot.herokuapp.com/view/guardian/typerighter

I think if Prout was making a POST request to checker.typerighter.gutools.co.uk, it would presumably get a fresh response from the cache, but it would probably also need to authenticate, for which in Prout there's no mechanism available at the moment.

### DIFF
--- a/.prout.json
+++ b/.prout.json
@@ -1,6 +1,5 @@
 {
   "checkpoints": {
-    "Checker": { "url": "https://checker.typerighter.gutools.co.uk/healthcheck", "overdue": "15M" },
     "Rule Manager": { "url": "https://manager.typerighter.gutools.co.uk/healthcheck", "overdue": "15M" }
   }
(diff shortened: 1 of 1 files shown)


jonathonherbert:
Thanks for this, @rtyley! I don't think we should be caching at all – https://github.com/guardian/typerighter/pull/468 explains why, and removes caching entirely, which should mean this is unnecessary ✨

PR #467, opened 2024-04-08
Add similarity ordering to rule search

jonathonherbert:
## What does this change?

I was searching for a dictionary rule, and realised that Typerighter's rule search is not very good. This PR adds similarity ordering to rule search.

Part of the reason is that we use a trigram index for comparing words, but do not order by similarity to the search term. Added here.

|Before|After|
|--|--|
|Screenshot 2024-04-08 at 09 40 25|Screenshot 2024-04-06 at 15 18 48|

## How to test

- The automated tests should pass. Existing ones have reasonable coverage of search functionality. There's a new one for relevance.
- Running locally or in CODE, observe the behaviour of search. It should favour text that most closely matches the rule type, pattern, description, or replacement.


rhystmills (approved):
Brilliant work.

Deployed to CODE and search is miles better. This has been bothering me also:
| Before | After |
| --- | --- |
|  |  |

PR #468, opened 2024-04-10
Don't cache the checker healthcheck endpoint

jonathonherbert:
## What does this change?

Don't use caching for the Checker service, to enable the lovely Prout-bot to do its work (#465.)

We use Cloudfront to improve service latency for global users by removing TCP slowstart and reducing TLS handshake latency – we don't care about caching content.

(Initially this PR just removed caching for the healthcheck endpoint (620eb3719971b45860a945ea17ac6794cad54a06), but on reflection I think removing caching entirely is more appropriate.)

## How to test

Deploy to CODE or PROD. You should no longer see `X-Cache: Hit from cloudfront` headers on requests.

### DIFF
--- a/cdk/lib/__snapshots__/index.test.ts.snap
+++ b/cdk/lib/__snapshots__/index.test.ts.snap
@@ -1767,39 +1767,6 @@ exports[`The typerighter stack matches the snapshot 1`] = `
       },
       "Type": "AWS::EC2::SecurityGroupEgress",
     },
-    "checkercloudfrontcachepolicyC5791389": {
-      "Properties": {
-        "CachePolicyConfig": {
-          "DefaultTTL": 86400,
-          "MaxTTL": 31536000,
-          "MinTTL": 0,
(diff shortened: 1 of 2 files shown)


simonbyford (commented):
This seems like a reasonable change. Unfortunately when I deployed the change to CODE, I was greeted by this error:

Screenshot 2024-07-31 at 16 17 50

https://checker.typerighter.code.dev-gutools.co.uk/rules

I appreciate it's been a while since you proposed this change. Do you remember seeing anything like this in your testing?

jonathonherbert:
Thanks for testing thoroughly! I'll revisit this PR soon.

jonathonherbert:
... for a given value of `soon` 😅  [76fef75](https://github.com/guardian/typerighter/pull/468/commits/76fef7583817d9e3ca99ad2dbd5dfd2fd9ba9279) should fix.

rhystmills (approved):
Looking good on CODE 👍

PR #469, opened 2024-04-10
Add filters to search

jonathonherbert:
## What does this change?

Adds search filters for tags and rule type.

filter-by-tag

## Implementation notes

In altering the search query for this I realised we didn't need the subquery – see [fb130e3](https://github.com/guardian/typerighter/pull/469/commits/fb130e39d2221e26c19ee9ca2e70fb640a642102). Minor complexity reduction.

## How to test

- The automated test should pass.
- Run locally or (better) in CODE, and have a play with the filters. They should behave as expected in combination with the text search.

## What about a filter for publication status? Or source?

Wouldn't that be nice? PRs welcome!

There's a question of how we manage an expanding set of filters.

### DIFF
--- a/apps/rule-manager/app/controllers/RulesController.scala
+++ b/apps/rule-manager/app/controllers/RulesController.scala
@@ -40,7 +40,7 @@ class RulesController(
             .map(toFormError("refresh-sheet"))
           _ <- RuleManager.destructivelyPublishRules(dbRules, bucketRuleResource)
         } yield {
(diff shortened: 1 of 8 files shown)


paperboyo:
💎 

> There's a question of how we manage an expanding set of filters.

🍟 ?

rhystmills (approved):
Delightful feature, especially handy for searching Regex and LanguageTool rules with all the Collins Dictionary rules in the table 👍

PR #471, opened 2024-06-06
Fix pagination results

jonathonherbert:
## What does this change?

Adds a test and a fix for the odd pagination results patched in #470.

Not quite sure what was causing the original discrepancy, but we were running two queries, one for the data, and one for the count, which made the inconsistency possible.

This PR moves the count columns into the original query, which should remove that inconsistency.

## How to test

- Running locally or in CODE, use the search bar and observe the paginationresults given back by the API. They should match the data returned.


jonathonherbert:
@Fweddi great suggestion, done in [67aaf1f](https://github.com/guardian/typerighter/pull/471/commits/67aaf1ffa57c6159a9bb37ec78acbbf2a8b99d8f) and pagination looks good.

Fweddi (approved):

PR #472, opened 2024-06-26
Bump Scala dependencies to solve vulnerabilities

rhystmills:
## What does this change?

This bumps some Scala package versions in Typerighter to resolve some high priority vulnerabilities. According to Snyk, this drops the number of high priority backend  vulnerabilities from 7 to 1. 

Snyk runs for commits in this PR, from newest to oldest:

image

The remaining vulnerability seems to be unsolved in the latest version of Play.

## How to test

1. Does the app compile, and do the tests pass? 
2. Deploy to CODE. Does the app work as expected? (Does the rule manager work, can you get matches in CODE Composer?


rhystmills:
(Tested Rule Manager and Checker on CODE, also logs - all working well)

PR #473, opened 2024-06-26
Bump @guardian/cdk and jest to solve vulnerabilities

rhystmills:
## What does this change?

Bumps @guardian/cdk and jest to their latest versions to solve vulnerabilities. `GuDatabaseInstance `now requires a `devXBackups` argument so we don't need to add a tag ourselves.

I don't think any of the other breaking changes to `@guardian/cdk` affect Typerighter, though that library now recommends ALBs over ELBs.

## How to test

1. Does it build and run?
2. Deploy to CODE. Do both apps and the Composer integration work as expected?
3. Do the Jest tests run and pass?


phillipbarron (approved):
Works in CODE, both manager and Composer integration & test pass

PR #483, opened 2024-11-05
Read and write core LT rules

jonathonherbert:
## What does this change?

Adds proper support for reading and writing `languageToolCore` rules — rules which come bundled with LanguageTool. They are enabled by adding a rule with a LanguageTool ID. The total list of rules is available in a [reference spreadsheet](https://docs.google.com/spreadsheets/d/1bLCD7CrGMDAuuk76e0MTF1-DUNgmXrvTc1d5n2UMZR8/edit?gid=1482142974#gid=1482142974) — many (most?) are also available in the [LanguageTool XML for `en`](https://raw.githubusercontent.com/languagetool-org/languagetool/refs/heads/master/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml) and [en-GB](https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/en-GB/grammar.xml), although I think some rules are also defined exclusively [in code](https://github.com/languagetool-org/languagetool/tree/master/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en), too.

Support for reading and writing rules means:

- Listing the ruleType property in the table, with filtering/sorting
- Copying the 'ID' property into the 'description' for existing rules (there are only a dozen or so in production), making them identifiable (at the moment, there are blank rows for these rules, which makes them impossible to identify without selecting them.)

built-in rules

- Permitting create and update operations for these rules in the relevant form. Note that I've autopopulated the ID field with 'CHANGE_ME' to avoid a UUID appearing when the rule is first created on the server. (We could also leave it blank?)

create

## How to test

- The automated tests should pass. I've added tests for create and update ops.
- Try working through the list of features above. Everything should work as expected.

More editorial control over these sorts of rules.

### DIFF
--- a/apps/rule-manager/app/db/DbRuleDraft.scala
+++ b/apps/rule-manager/app/db/DbRuleDraft.scala
@@ -10,6 +10,8 @@ import scalikejdbc._
 
 import java.time.OffsetDateTime
 import scala.util.{Failure, Success, Try}
+import utils.StringHelpers
+import service.RuleManager.RuleType
 
 case class DbRuleDraft(
     id: Option[Int],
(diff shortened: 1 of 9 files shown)


KaliedaRik (approved):
Have smoke tested the branch in my local environment, and also replicated the user journeys offered as examples in the PR discussion. Have also run the tests locally and seen all pass.

Happy to approve.

PR #490, opened 2025-01-16
Upgrade to panda v7

waisingyiu:
## What does this change?

This pull request upgrades Pan-domain-auth to v7 which supports key rotation.

I follow the guidance from https://github.com/guardian/pan-domain-authentication/issues/160 to make the following changes - 
1. bump the panda version to 7.0.0
2. create the `PanDomainAuthSettingsRefresher` instance using the `apply` of the companion object with `s3BucketLoader`, as recommended by the library.

I've also replaced the `setup-java` action with our `setup-scala` action in the CI github action to fix the build.

## How to test

The changes were deployed to `CODE` and tested in the following ways:

1. When I accessed the `Typerighter` without sign-in, we were redirected to Google login page and, upon sign-in, to `Typerigher`.  I could access `Typerighter` then.  I could also access Composer and Fronts tools without signing in again.

2. When I signed in via Fronts tools first, I could access `Typerighter` without signing in again.


davidfurey (approved):

jonathonherbert:
> Overdue on [Checker](https://checker.typerighter.gutools.co.uk/healthcheck) (merged by @waisingyiu 15 minutes and 3 seconds ago) What's gone wrong?

This is a bug in our CDN configuration: https://github.com/guardian/typerighter/pull/468, which needs some attention, aims to fix. So there's no action here.

PR #492, opened 2025-01-23
Bump languagetool to 6.4

rhystmills:
## What does this change?
This bumps [`languagetool`](https://www.google.com/search?q=languagetool+github&rlz=1C5CHFA_enGB1056GB1056&oq=languagetool+github&gs_lcrp=EgZjaHJvbWUyCQgAEEUYORiABDIHCAEQABiABDIHCAIQABiABDIHCAMQABiABDIHCAQQABiABDIGCAUQRRg8MgYIBhBFGDwyBggHEEUYPNIBCDI1ODJqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8) to version 6.4. There are a few high priority vulnerabilities in our sbt dependency, and some of them are sub-dependencies of `languagetool`.

The newest `languagetool` version [according to maven is 6.6](https://mvnrepository.com/artifact/org.languagetool/languagetool-core), but sbt wasn't able to find it. I've only bumped to Version `6.4` because `6.5` introduced a fiddly-looking error (though it compiles successfully) and I'm hoping that `6.4` will be sufficient to solve some of our dependency woes.

`languagetool` does not appear to be semantically versioned as there were breaking changes in `6.1` and `6.3` as well as `6.5`. I've had to make a couple of small changes to address the changes in the `6.1` and `6.3`, namely:
- Updating the grammar used in a test
- Providing a language to `getRules`

It's definitely worth testing this PR in CODE Composer and seeing if checks from our rule manage show up successfully!

I'd be especially keen to check some dictionary rules, because they rely on some slightly [arcane workarounds](https://github.com/guardian/typerighter/pull/449), and undocumented changes in the structure of `languagetool` could affect them.

## How to test

1. Deploy this branch to CODE.
2. Find a couple of rules representing each rule type in the [CODE rule manager](https://manager.typerighter.code.dev-gutools.co.uk/)(i.e. Dictionary, LT Built-In, Regex).
3. Test those rules in CODE Composer. Is the Checker service working as intended, and surfacing them in Composer? E.g. do you see a red or orange underline where you would expect to?

I can advise on any aspects of testing Typerighter if you're not familiar with the tool, please reach out if you require some help

### DIFF
--- a/apps/checker/test/scala/matchers/LanguageToolMatcherTest.scala
+++ b/apps/checker/test/scala/matchers/LanguageToolMatcherTest.scala
@@ -218,7 +218,7 @@ class LanguageToolMatcherTest extends AsyncFlatSpec with Matchers {
     val eventuallyMatches = instance.check(request)
 
     val expectedMatchMessages =
-      List("Did you mean <suggestion>fewer</suggestion>? The noun tests is countable.")
+      List("Did you mean <suggestion>fewer</suggestion>? The noun \"tests\" is countable.")
     val expectedMatchCategoryIds = List("GRAMMAR")
     eventuallyMatches map { matches =>
       matches.map(_.message) shouldBe expectedMatchMessages
--- a/apps/checker/app/matchers/LanguageToolMatcher.scala
+++ b/apps/checker/app/matchers/LanguageToolMatcher.scala
@@ -1,8 +1,10 @@
 package matchers
 
 import com.gu.typerighter.model.{Category, LTRule, LTRuleXML, RuleMatch}
+
 import java.io.File
 import org.languagetool._
+import org.languagetool.language.{BritishEnglish}
 import org.languagetool.rules.spelling.morfologik.suggestions_ordering.SuggestionsOrdererConfig
 import org.languagetool.rules.{Rule => LanguageToolRule}
 import play.api.Logging
@@ -13,7 +15,8 @@ import scala.concurrent.{ExecutionContext, Future}
 import org.languagetool.rules.patterns.PatternRuleLoader
 import org.languagetool.rules.patterns.AbstractPatternRule
 import utils.{Matcher, MatcherCompanion}
-import scala.xml.{XML, Attribute, Null, Text}
+
+import scala.xml.{Attribute, Null, Text, XML}
 import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
@@ -111,7 +114,12 @@ class LanguageToolFactory(
         val loader = new PatternRuleLoader()
         getXMLStreamFromLTRules(rules) flatMap { xmlStream =>
           {
-            Try(loader.getRules(xmlStream, "languagetool-generated-xml").asScala.toList)
+            Try(
+              loader
+                .getRules(xmlStream, "languagetool-generated-xml", new BritishEnglish())
+                .asScala
+                .toList
+            )
           }
         }
       }
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ ThisBuild / scalacOptions := Seq(
 // See https://support.snyk.io/hc/en-us/articles/9590215676189-Deeply-nested-Scala-projects-have-dependencies-truncated
 ThisBuild / asciiGraphWidth := 999999999
 
-val languageToolVersion = "6.0"
+val languageToolVersion = "6.1"
 val awsSdkVersion = "1.12.749"
 val capiModelsVersion = "17.5.1"
 val capiClientVersion = "19.2.1"
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ ThisBuild / scalacOptions := Seq(
 // See https://support.snyk.io/hc/en-us/articles/9590215676189-Deeply-nested-Scala-projects-have-dependencies-truncated
 ThisBuild / asciiGraphWidth := 999999999
 
-val languageToolVersion = "6.3"
+val languageToolVersion = "6.4"
 val awsSdkVersion = "1.12.749"
 val capiModelsVersion = "17.5.1"
 val capiClientVersion = "19.2.1"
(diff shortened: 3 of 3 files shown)


emdash-ie:
> The newest `languagetool` version [according to maven is 6.6](https://mvnrepository.com/artifact/org.languagetool/languagetool-core), but sbt wasn't able to find it.

It looks like this is because the 6.6 versions are in [a new dedicated repository](https://mvnrepository.com/repos/languagetool-os-release), based on [the mvnrepository page for 6.6.9](https://mvnrepository.com/artifact/org.languagetool/languagetool-core/6.6.9). Maven Central [agrees with sbt](https://central.sonatype.com/artifact/org.languagetool/languagetool-core) that the newest version there is 6.5.

Do we want to use languagetool’s apparent dedicated repository? Is that a thing we generally do? (I suppose it means we don’t get the same third-party checking we get from Maven Central, but I don’t really know the details of what we get from them.) I tried looking at [their developer docs](https://dev.languagetool.org/java-api), but didn’t find any explanation there about the new repository.

apps/checker/test/scala/matchers/LanguageToolMatcherTest.scala
```diff
@@ -218,7 +218,7 @@ class LanguageToolMatcherTest extends AsyncFlatSpec with Matchers {
     val eventuallyMatches = instance.check(request)
 
     val expectedMatchMessages =
-      List("Did you mean <suggestion>fewer</suggestion>? The noun tests is countable.")
+      List("Did you mean <suggestion>fewer</suggestion>? The noun \"tests\" is countable.")
```
emdash-ie, line 221: I’m surprised this match text requires quotes but the one below (`The noun mistakes…`) doesn’t. What’s the difference?

jonathonherbert:
A quick note to say that we rely on some languagetool rules which may have changed behaviour between releases. You can see the full list by filtering rules by 'LT Built-in' in the rule manager — 
Screenshot 2025-02-03 at 16 28 51

There's a script [here](https://github.com/guardian/typerighter/blob/main/script/js/compare-rule-xml.js) to check the difference between old and new rules at `compare-rule-xml` which might be worth running to see if any of the rule definition XML has changed. The script should document itself, but I don't think anyone's used it yet, so super happy to pair on this if it's not clear what to do with it.

I don't think this is necessarily blocking, as our users will let us know if something's obviously not right, but it might be nice to get ahead of any changes before they land in PROD.

emdash-ie:
> There's a script [here](https://github.com/guardian/typerighter/blob/main/script/js/compare-rule-xml.js) to check the difference between old and new rules at `compare-rule-xml` which might be worth running to see if any of the rule definition XML has changed. The script should document itself, but I don't think anyone's used it yet, so super happy to pair on this if it's not clear what to do with it.

Thanks, will test this

emdash-ie:
I have [deployed this to CODE](https://riffraff.gutools.co.uk/deployment/view/664e3ad3-ba29-45d2-be14-39dd55efdbf6) and tested it in [this article](https://composer.code.dev-gutools.co.uk/content/67a0b2b48f08012752b767be). I've checked a number of built-in, regex, and dictionary rules, which appear to work correctly (though it would be good to take a look with someone who's more familiar with typerighter).

I have had a hard time triggering some specific dictionary rules, e.g. 1,2-dihydroxypropane. But several appear to have been triggered correctly, so maybe this is fine.

emdash-ie:
I ran the script and got a slightly inconclusive answer. I downloaded the english grammer file the script points to from [v6.0](https://github.com/languagetool-org/languagetool/blob/v6.0/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml) and from [v6.4](https://github.com/languagetool-org/languagetool/blob/v6.4/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml), wrote  a rules.txt based on the LT built-in rules listed in [typerighter-CODE](https://manager.typerighter.code.dev-gutools.co.uk/), and then ran the script:

```bash
  yarn
  node compare-rule-xml.js grammar-6.0.xml grammar-6.4.xml rules.txt
```

Which gave this output:

> Checking PUBIC_X ...
> PUBIC_X is unchanged
> Checking IN_PRINCIPAL ...
> IN_PRINCIPAL is unchanged
> Checking UPPERCASE_SENTENCE_START ...
> UPPERCASE_SENTENCE_START not found in either file
> Checking WHITESPACE_RULE ...
> WHITESPACE_RULE not found in either file
> Checking SENTENCE_WHITESPACE ...
> SENTENCE_WHITESPACE not found in either file
> Checking NO_SPACE_CLOSING_QUOTE ...
> NO_SPACE_CLOSING_QUOTE is unchanged
> Checking CURRENCY_SPACE ...
> CURRENCY_SPACE is unchanged
> Checking DOUBLE_PUNCTUATION ...
> DOUBLE_PUNCTUATION not found in either file
> Checking  ...
>  not found in either file

A few of the rules (UPPERCASE_SENTENCE_START, WHITESPACE_RULE, SENTENCE_WHITESPACE, and DOUBLE_PUNCTUATION) were not found in either file, which suggests they might be somewhere else? I haven't been able to locate a couple of them in xml files by searching the languagetool repository (they do appear in java files, which I assume this script can't check?).

emdash-ie:
I think I have been able to test each of those rules in [my test article](https://composer.code.dev-gutools.co.uk/content/67a0b2b48f08012752b767be) and have seen them correctly flagged.

emdash-ie (approved):
This looks good to me, but it would be good to discuss on a call before merging, to make sure I have tested it well enough.

jonathonherbert:
> A few of the rules (UPPERCASE_SENTENCE_START, WHITESPACE_RULE, SENTENCE_WHITESPACE, and DOUBLE_PUNCTUATION) were not found in either file, which suggests they might be somewhere else? I haven't been able to locate a couple of them in xml files by searching the languagetool repository (they do appear in java files, which I assume this script can't check?).

That's spot on, @emdash-ie, some rules are defined as Java classes in the library. Thanks for checking the others — that gives us some confidence they're working correctly.

PR #493, opened 2025-01-27
Add script to dump contents of rule DB to CSV

jonathonherbert:
## What does this change?

Adds a script to dump the contents of the rule DB to CSV.

## How to test

Run the script, amending the configuration as necessary if you'd like to pull rules from CODE or PROD.

### DIFF
--- a/script/js/dump-db.ts
+++ b/script/js/dump-db.ts
@@ -1,0 +1,117 @@
+import { Pool } from 'pg';
+import { createObjectCsvWriter } from 'csv-writer';
+import { format } from 'date-fns';
+
+const dbConfig = {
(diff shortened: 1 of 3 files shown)


jonathonherbert:
That would be best, and thanks for reminding me about the export endpoint! This code has the merit of existing, and I'm unsure when we'll find the time to do that; we could merge this with the intent to add an endpoint later, or close.

paperboyo:
Almost as if `/spreadsheet` should be a [standard route](https://github.com/guardian/tagmanager/pull/360) for everything with a db… ;-)

davidfurey (approved):

PR #498, opened 2025-03-06
Snappier search

jonathonherbert:
## What does this change?

Makes a few changes to make search in our rule manager snappier. It's quite a bit slower than it was when we first implemented it, because our corpus has grown from 16,000 to ~340,000 rules.

There are a few differences in the DB and the frontend:
- There are indices for similarity search and common ordering patterns
- The main search query is in a subselect to avoid joins and grouping on the search query where possible 
- Rule type is no longer in the full-text search query/index, as we can now search by rule type explicitly in UI. This improves results a bit, too, as the rule type does not factor into the similarity distance, and rule type does not contain semantics that are unique to the rule
- Implement request cancellation on the client, as long queries were overlapping inflight and occasionally causing results to overlap, a disconcerting bug
- Remove the default 'pattern' sort on the client to avoid sorting by pattern when we are subsequently searching for strings — similarity should take precedence, again improving results (try searching for e.g. `Labour` on this branch and compare to prod). To provide a deterministic result, we default to sort by `updated_at` rules when no other sort is provided on the server — probably nice to see the most recently changed rules first?

|Before|After|
|--|--|
|slow-tr|fast-tr|

There are a few different sorts of changes in this PR, and it might be useful to be review per-commit. Although the LoC is relatively small and the features group together nicely, I appreciate there's a fair amount to examine. Happy to pair to review, or split this PR into smaller pieces if that'd be useful.

## How to test

The easier way is probably to give it a spin in CODE. Does the search behave as expected? Does it feel quicker than PROD, or the gif above?

Less pain searching for rules in Typerighter.


apps/rule-manager/app/db/DbRuleDraft.scala
```diff
@@ -280,49 +271,77 @@ object DbRuleDraft extends SQLSyntaxSupport[DbRuleDraft] {
...
     }
 
-    val data = sql"""
+    val countStmt = if (searchClause.nonEmpty) {
+      sqls"COUNT(${rd.id}) OVER () AS rule_count"
+    } else {
+      sqls"""
+        (SELECT n_live_tup
+          FROM pg_stat_all_tables
+        WHERE relname = ${DbRuleDraft.tableName}) as rule_count
```
itsibitzi, line 299: fwiw this will give an estimated figure of the number of live tuples which could lead to surprising results under certain circumstances.

Probably won't be a problem for you since I suspect the number of writes you receive is low.

jonathonherbert: Thanks for pointing this out, I meant to add a comment here — [f025fc5](https://github.com/guardian/typerighter/pull/498/commits/f025fc5233872c13195895029245d79408989bb6)

apps/rule-manager/conf/evolutions/default/17.sql
```diff
@@ -0,0 +1,21 @@
+-- !Ups
+
+CREATE INDEX rules_draft_free_text_search_idx
+    ON public.rules_draft
+        USING gist ((coalesce(rules_draft.pattern, '') ||
```
andrew-nowak, line 5: worth trying this with a `gin` index if you haven't already - I believe in general it's faster, at the expense of slower updates, and some disk space https://stackoverflow.com/a/28976555 , and also it looks like [some trgm queries won't use a `gin` ](https://www.postgresql.org/docs/17/pgtrgm.html#PGTRGM-INDEX:~:text=t%20AS%20dist%0A%20%20FROM%20test_trgm%0A%20%20ORDER%20BY%20dist%20LIMIT%2010%3B-,This%20can%20be%20implemented%20quite%20efficiently%20by%20GiST%20indexes%2C%20but%20not%20by%20GIN%20indexes,-.)

jonathonherbert: GIN was my first choice! When I was testing, I was pretty confused as to why the index wasn't being hit until I read [this](https://stackoverflow.com/a/34641164) and removed the `ORDER BY` clause! When I use GIST, I can see the index being hit in the query we generate when I `EXPLAIN`. So I went with GIST.

The official docs were quite vague on what I should expect — the link above doesn't clearly scope which statements `This can be implemented quite efficiently by GiST indexes, but not by GIN indexes.` applies to — and it seems an odd shortcoming, as sorting by similarity is important in providing good quality results: perhaps I'm missing something here?

jonathonherbert: (Sidenote: I'm really keen to understand this space better, b/c many of our tools would benefit from good quality free text search and a database that can provide data integrity guarantees)

andrew-nowak: heartbreaking. I wonder if there's some way to change the query to nudge postgres to use the index, but it's probably not worth the extra investigation since the gist index is already looking like a pretty good improvement

itsibitzi: There's `pg_hint_plan` which IIRC is available in RDS but it's a pain in the arse.

As I understand it GIN indexes don't maintain ordering so can't satisfy the `ORDER BY` clause, GiST indexes do have ordering tho. 

Use of indexes will as always depend on the specific query and the stats Postgres generates. :)

paperboyo: il_300x300 3959367342_pwp0

apps/rule-manager/conf/evolutions/default/17.sql
```diff
@@ -0,0 +1,21 @@
...
+                    coalesce(rules_draft.description, '') ||
+                    coalesce(rules_draft.replacement, '')) gist_trgm_ops);
+
+-- We order by updated_at by default - this index speeds up the initial call to display rules
+CREATE INDEX rules_draft_updated_at ON rules_draft(updated_at DESC);
+
+-- `left` is used to index on a substring, as patterns can be very large, and there is a limit
+-- on the size of the data that can cause write errors (https://stackoverflow.com/a/70124913)
+CREATE INDEX rules_draft_pattern_idx_asc ON rules_draft(left(pattern, 20));
+CREATE INDEX rules_draft_pattern_idx_desc ON rules_draft(left(pattern, 20) DESC);
```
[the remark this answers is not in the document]
jonathonherbert: In practice locally and on CODE, the indices took ~2s to create on a tool with minimal write volume, so I didn't think too hard about this — but it's a good point that CONCURRENTLY is probably a sensible default! 19cc3a52d40c9d8f0b47a71012ae7a34d50186b7

apps/rule-manager/conf/evolutions/default/17.sql
```diff
@@ -0,0 +1,21 @@
...
+
+-- !Downs
+
+DROP INDEX rules_draft_free_text_search_idx;
+DROP INDEX rules_draft_pattern_idx_asc;
+DROP INDEX rules_draft_pattern_idx_desc;
```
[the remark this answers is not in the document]
jonathonherbert: Yes! Thanks for spotting. e5fb43021ad7a00be3af69e459f3621e973ba28a

jonathonherbert:
(note to self: [e5fb430](https://github.com/guardian/typerighter/pull/498/commits/e5fb43021ad7a00be3af69e459f3621e973ba28a) and [19cc3a5](https://github.com/guardian/typerighter/pull/498/commits/19cc3a52d40c9d8f0b47a71012ae7a34d50186b7) will cause a migration conflict in CODE, resolve before merging)

apps/rule-manager/conf/evolutions/default/17.sql
jonathonherbert: I've been over my previous queries, and my results when attempting to use GIN were missing the index not because of an errant `ORDER BY`, but because the query string was too short! A comparison string with one or two  characters, used in previous attempts to exercise inefficiencies in the count and join parts of the query, isn't enough to persuade PG to use the index 🤦 reasonable as the index is composed of trigrams?

Not thinking that PG would care about the size of the string in planning its query, I misattributed the index miss to the `ORDER BY` clause. But although the docs suggest that GIN can't be used for ordering, either the query for large enough search terms is so much faster that if this is true, it doesn't seem to matter (stats in MS taken from `EXPLAIN ANALYZE` preceding queries taken direct from frontend searches):

branch | [no query] | `a` (misses index) | `example`
-- | -- | -- | --
main 1 | 832 | 1024 | 78
main 2 | 815 | 1044 | 74
main 3 | 806 | 1026 | 75
gist 1 | 17 | 465 | 17
gist 2 | 14 | 452 | 19
gist 3 | 13 | 468 | 17
gin 1 | 16 | 462 | 1.5
gin 2 | 13 | 472 | 1.2
gin 3 | 16 | 461 | 3.7

We can see one (and two) char queries missing the index and taking much longer. Queries this short unlikely to be useful in themselves, more likely to be triggered by an inadvertent debounce as the user is typing, so we could prevent searches for <3 chars.

GIN index in [1ecbb21](https://github.com/guardian/typerighter/pull/498/commits/1ecbb214015d65b15bdd268b9403e271482ee7b0). Happy to see PG being super super fast.

apps/rule-manager/conf/evolutions/default/17.sql
```diff
@@ -0,0 +1,22 @@
+-- !Ups
+
+CREATE INDEX CONCURRENTLY rules_draft_free_text_search_idx
+    ON public.rules_draft
+        USING gist ((coalesce(rules_draft.pattern, '') ||
+                    coalesce(rules_draft.description, '') ||
+                    coalesce(rules_draft.replacement, '')) gist_trgm_ops);
```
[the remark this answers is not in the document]
jonathonherbert: 💡 thanks for spotting! [abdae02](https://github.com/guardian/typerighter/pull/498/commits/abdae02e9ed03c61250154354607982cbe37e941)

jonathonherbert:
Latest changes tested in CODE. Even snappier.

andrew-nowak (approved):
looks good to me - great speedup!

PR #527, opened 2025-10-14
Remove content-api-client-aws, as it is unused

jonathonherbert:
## What does this change?

Remove content-api-client-aws, as it is unused.

## How to test

Project should compile.


emdash-ie (approved):
Looks good! I think this dependency was mistakenly added in [this PR from 2020](https://github.com/guardian/typerighter/pull/36) and kept around since?

Also, the dependency is listed a second time on line 141: I think we should remove it there too before merging, so I’ll push a commit to do that.

jonathonherbert:
Oh, thank you for spotting that! 🙏

PR #529, opened 2025-10-22
Upgrade AWS SDK to v2

waisingyiu:
## What does this change?

AWS SDK for Java v1 will be EOL on 31st December 2025. It has been in Maintenance mode since [31st July 2024](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/).

The pull request upgrades its AWS SDK to the latest v2 version.  It also bumps the "pan-domain-authentication" and "permission client libraries" to their latest version because they have dependency on AWS SDK too.

## How to test

On typerigher rule management, we can
- see the existing rules
- update an existing rule
- add a new rule
- after adding a rule, we can see a data point in the cloudwatch metric "RulesInjested"

On Composer, we can
- run checker on text block with the new rule added in the management

Migrated to AWS SDK v2, and we don't have this service listed under our team in the AWS SDK v1 dashboard

The risk should be small, as we tested it on CODE.

## Images

Run checker in Composer with a new rule added in Typerigher rule manager:
Screenshot 2025-10-23 at 12 40 06

The cloudwatch metric after adding a new rule:
Screenshot 2025-10-23 at 14 25 59

### DIFF
--- a/apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
+++ b/apps/common-lib/src/main/scala/com/gu/typerighter/rules/BucketRuleResource.scala
@@ -4,7 +4,7 @@ import software.amazon.awssdk.services.s3.S3Client
 import software.amazon.awssdk.services.s3.model.{
   PutObjectRequest,
   GetObjectRequest,
-  GetObjectAttributesRequest
+  HeadObjectRequest
 }
 import software.amazon.awssdk.core.sync.RequestBody
 import com.gu.typerighter.lib.JsonHelpers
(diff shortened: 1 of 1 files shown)


apps/common-lib/src/main/scala/com/gu/typerighter/controllers/PandaAuthController.scala
```diff
@@ -24,7 +24,6 @@ abstract class PandaAuthController(
     with HMACAuthActions
     with BaseController
     with Loggable {
-  override val apiGracePeriod: Long = 2 * 60 * 60 * 1000 // two hours worth of milliseconds
```
waisingyiu: This value is removed in PR https://github.com/guardian/pan-domain-authentication/pull/185 because
```
We want to move towards enforcing this pattern consistently across tools, so we are setting the default within the library to 24 hours.
```

PR #530, opened 2025-10-29
Upgrade languagetool to fix io.grpc:grpc-netty-shaded vulnerability

LAKSHMIRPILLAI:
## What does this change?

A high-severity dependabot alert was raised about the library "io.grpc:grpc-netty-shaded", which is pulled in as a transitive dependency through LanguageTool.

The pull request bumps the LanguageTool to the latest version 6.7, which pulls in "io.grpc:grpc-netty-shaded" v1.75 that fixes the vulnerability.

## How to test

All the test cases passed.  We can do a smoke test on the typerighter manager and on the checker in Composer.
- add and publish new rules in manager
- modify existing rules in manager
- run checker in Composer with words that hit the new rules added in manager 

The high-severity dependabot alert is cleared.

The risk should not be high, as it mainly changes the version of its dependencies and the Java runtime.  Some minor changes in the program to fix the test cases, but it is not expected to cause any changes to the program behaviour.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -19,7 +19,7 @@ deployments:
       prependStackToCloudFormationStackName: false
       amiParametersToTags:
         AMITyperighterchecker:
-          Recipe: editorial-tools-jammy-java17
+          Recipe: editorial-tools-jammy-java17-ngrams
           BuiltBy: amigo
         AMITyperighterrulemanager:
           Recipe: editorial-tools-jammy-java17


apps/checker/app/services/collins/CollinsEnglish.java
```diff
@@ -16,6 +16,11 @@
 import java.util.ResourceBundle;
 
 public class CollinsEnglish extends BritishEnglish {
+
+    public CollinsEnglish() {
+        super(false);
```
waisingyiu: The argument to the `super()` isn't really used.  The new version of LanguageTool creates another constructor for subclasses to instantiate directly, and the argument is just intended to make it different from the default constructor.

dblatcher (approved):
tested on code. Looks good to me!

PR #540, opened 2025-12-15
Add pixel tracking to typerighter

waisingyiu:
## What does this change?

This PR adds the [Tools Audit tracking pixel](https://github.com/guardian/editorial-tools-user-telemetry-service?tab=readme-ov-file#tracking-pixel) to allow us to understand more how the facia tool is being used by different users. The results will be visible in the [tools audit dashboard](https://metrics.gutools.co.uk/d/aegzv59r9eghsd/tool-audit?orgId=1&var-domain=birthdays.gutools.co.uk&var-path=All&var-interval=1d&var-chosen_apps=All).

This PR adds the pixel tracking only to the pages in the frontend that are directly accessed by users.  It does not cover API usage.

We use `typerighter-manager` and `typerighter-checker` as the app names of the two services in Typerighter.

## How to test

Open Typerighter manager and checker with the web inspector tool
- we can see the HTTP requests to the telemetry service with `guardian-tool-accessed` path when we switch pages.

We can see the usage data on the Tools Audit dashboard.

Very low as we do not change the application logic in this PR.

### DIFF
--- a/apps/rule-manager/client/src/ts/components/layout/Page.tsx
+++ b/apps/rule-manager/client/src/ts/components/layout/Page.tsx
@@ -59,7 +59,11 @@ export type RouteHandle = {
 	useAsPageTitle?: boolean;
 };
 
-function usePixelTracking() {
+const PageContent: React.FC = () => {
+	const maybeMatch = useMatches()
+		.filter((match) => (match.handle as RouteHandle).useAsPageTitle !== false)
+		.pop();
+	const name = maybeMatch ? getNameFromRouteMatch(maybeMatch) : 'Unknown route';
 	const location = useLocation();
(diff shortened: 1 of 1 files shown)


waisingyiu:
> Tested locally and on CODE.
> 
> There are some CSP issues on the `checker` side which prevent the tracking requests being made:
> 
> There is also some stray braces at the top of the page:
> 

Thanks Sam for spotting these!  I've added the gutools domain (for CODE and for PROD) to the CSP related config to allow loading resources from telemetry service.  I've also fixed the stray braces at the checker page.

SHession (approved):
Tested again on CODE. I can see network requests being made and then surfaced in our telemetry data. LGTM!

PR #631, opened 2026-06-04
Fix vulnerability in thrift library

waisingyiu:
## What does this change?
The pull request adds the thrift library as a direct dependency to bump it to v0.23.0 which fixes a high severity vulnerability:
- https://github.com/guardian/typerighter/security/dependabot/353

The libraries that pull in Thrift still depend on the vulnerable version of Thrift so we cannot bump them to fix it.

## How has this change been tested?

Deploy to CODE and run a smoke test, creating a new rule and running a checker in Composer.

Run "sbt dependencyTree" to confirm the updated version of Thrift library is used.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -81,7 +81,8 @@ val commonSettings = Seq(
     // The jackson-module-scala version below must be kept in sync with the
     // transitive dependency on jackson-databind introduced by our AWS
     // dependencies.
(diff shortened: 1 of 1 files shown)


emdash-ie (approved):
Looks good! I can see you’ve deployed to CODE and presumably tested and checked for error logs.

SHession (approved):
I've done some basic checks on CODE. LGTM!

